PR #21, opened 2026-03-26
Fire `session_start` in subagent handlers

### LINKED ISSUE #20: Extension `session_start` handlers never fire in subagent sessions
_The below was written from an agent on a work computer so the libraries referenced may not be generally applicable. However I think the problem would remain in any other extension that has anything in `sesson_start` events. From what I can see, `session_start` is a key lifecycle event inside Pi itself so it's probably correct to propagate that to subagents._

# Bug report

## Summary

Subagent sessions created by `@tintinweb/pi-subagents` never call `session.bindExtensions()`, which means the `session_start` event is never emitted. Extensions that rely on `session_start` for initialization (e.g. loading credentials, setting up state) silently fail — their tools are registered and visible, but non-functional at runtime.

## Version

`@tintinweb/pi-subagents` v0.5.1

## Reproduction

Any extension that initializes state in a `session_start` handler will fail in a subagent. The Slack extension (`shop-pi-fy/extensions/slack`) is a concrete example:

```typescript
// In the extension:
pi.on("session_start", async (_event, ctx) => {
  loadCredsFromDisk(ctx);  // loads token from ~/.config/slack-mcp/credentials.json
});

pi.registerTool({
  name: "slack_search",
  async execute(/* ... */) {
    const c = await requireCredsAsync(); // throws "Slack not connected" because creds is null
    // ...
  },
});
```

**Parent session:** Works perfectly — `session_start` fires, creds load, tools work.

**Subagent session:** Tool is registered and visible, but every call throws `"Slack not connected"` because `session_start` never fired and credentials were never loaded.

### Steps to reproduce

```
# In a pi session with the Slack extension loaded:
Agent(
  subagent_type: "general-purpose",
  model: "anthropic/claude-haiku-4-5",
  prompt: "Call slack_search with query 'test' and count 1. Report the result.",
  max_turns: 2,
)
```

**Expected:** Slack search returns results (credentials loaded from disk).
**Actual:** `Error: Slack not connected. Run /slack-auth ...`

We confirmed this is NOT a credentials/network/path issue:
- The creds file exists and is readable from the subagent (`~/.config/slack-mcp/credentials.json`)
- `curl` to `slack.com/api/auth.test` with those creds succeeds from the subagent
- `HOME` and `PATH` are identical between parent and subagent
- The `slack_search` tool IS registered (the extension loaded and `registerTool` ran)

## Root cause

In `agent-runner.ts`, the subagent session is created and used like this:

```typescript
// agent-runner.ts, lines ~255-268
const { session } = await createAgentSession(sessionOpts);

// ... tool filtering ...

// Directly prompts without binding extensions:
await session.prompt(effectivePrompt);  // line ~346
```

`bindExtensions()` is never called. In pi's `AgentSession`, `session_start` is emitted **only** inside `bindExtensions()`:

```typescript
// pi-coding-agent/src/core/agent-session.ts
async bindExtensions(bindings: ExtensionBindings): Promise<void> {
    // ...
    if (this._extensionRunner) {
        this._applyExtensionBindings(this._extensionRunner);
        await this._extensionRunner.emit({ type: "session_start" });  // <-- only place this fires
        await this.extendResourcesFromExtensions("startup");
    }
}
```

In normal pi operation, `bindExtensions()` is called by the mode layer (interactive, print, or rpc) before any prompts are sent. The subagent runner skips this step.

### Call chain comparison

**Normal pi session:**
`createAgentSession()` → `mode.bindExtensions()` → `session_start` emitted → `session.prompt()`

**Subagent session:**
`createAgentSession()` → `session.prompt()` ← **`bindExtensions()` never called, `session_start` never fires**

## Impact

Any extension that uses `session_start` for initialization is broken in subagents. This includes:
- **Slack extension** — credential loading
- Potentially any extension that sets up auth, state, or connections in `session_start`

The failure is silent — tools appear available (they're registered during extension factory execution), but fail at runtime when they try to use uninitialized state.

## Suggested fix

Call `session.bindExtensions()` in `agent-runner.ts` after `createAgentSession()` and before `session.prompt()`. The bindings can be minimal stubs since subagents don't have interactive UI:

```typescript
const { session } = await createAgentSession(sessionOpts);

// Bind extensions to fire session_start and initialize extension state
await session.bindExtensions({
  onNotify: () => {},      // no-op for subagents
  onSetStatus: () => {},   // no-op for subagents
  onShutdown: () => {},
  onError: () => {},
});

// ... tool filtering, then prompt ...
await session.prompt(effectivePrompt);
```

Alternatively, if `bindExtensions` requires UI bindings that are hard to stub, consider emitting `session_start` directly after session creation in the runner.

bradwright:
Extensions which depend on the `session_start` event for functionality like authentication were not successfully triggering from `Agent` toolcalls as the event was not fired.

Discussion of the bug specifics are in issue #20 

Fixes #20

### DIFF
--- a/src/agent-runner.ts
+++ b/src/agent-runner.ts
@@ -292,6 +292,19 @@ export async function runAgent(
     session.setActiveToolsByName(activeTools);
   }
 
+  // Bind extensions so that session_start fires and extensions can initialize
+  // (e.g. loading credentials, setting up state). Placed after tool filtering
+  // so extension-provided skills/prompts from extendResourcesFromExtensions()
+  // respect the active tool set. All ExtensionBindings fields are optional.
+  await session.bindExtensions({
+    onError: (err) => {
+      options.onToolActivity?.({
+        type: "end",
+        toolName: `extension-error:${err.extensionPath}`,
+      });
+    },
+  });
+
   options.onSessionCreated?.(session);
 
   // Track turns for graceful max_turns enforcement
--- a/test/agent-runner.test.ts
+++ b/test/agent-runner.test.ts
@@ -76,6 +76,7 @@ function createSession(finalText: string) {
     steer: vi.fn(),
     getActiveToolNames: vi.fn(() => ["read"]),
     setActiveToolsByName: vi.fn(),
+    bindExtensions: vi.fn(async () => {}),
   };
   return { session, listeners };
 }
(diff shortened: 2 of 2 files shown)


tintinweb:
`+ @tintinweb/pi-subagents@0.5.2`

thanks @bradwright 🤗

PR #23, opened 2026-04-11
fix: respect PI_CODING_AGENT_DIR env var

Amolith:
Add `globalAgentDir()` helper that checks `$PI_CODING_AGENT_DIR` env var [already used by Pi](https://github.com/badlogic/pi-mono/tree/main/packages/coding-agent#environment-variables) before falling back to `~/.pi/agent/`.


Amolith:
Needed to polish a bit more, sorry for not opening as a draft to begin with!

tintinweb:
thanks for the PR. went with a slightly different fix via upstream `getAgentDir()` see #35

Amolith:
Thanks so much for getting it fixed! This had been on my list to get back to, but I hadn't had the time yet 😅

PR #28, opened 2026-04-21
fix: migrate to pi 0.68.0 breaking changes

Nachompiras:
- createAgentSession tools now expects string[] instead of Tool[]
  - replaced getToolsForType(cwd) with getToolNamesForType() returning string[]
  - replaced getMemoryTools(cwd, set) with getMemoryToolNames(set) returning string[]
  - replaced getReadOnlyMemoryTools(cwd, set) with getReadOnlyMemoryToolNames(set)
  - removed cwd-bound tool factory usage in agent-types.ts

- DefaultResourceLoader now requires agentDir (mandatory)
  - import and pass getAgentDir() explicitly

- SettingsManager.create() now requires cwd as first argument
  - pass effectiveCwd and agentDir explicitly

- session_switch event renamed to session_before_switch in ExtensionAPI


vekexasia:
:+1:

PR #29, opened 2026-04-21
Pass agentDir into subagent session setup

antikythera-technologies:
## Summary
- pass the resolved `agentDir` into the subagent resource loader
- create the subagent settings manager with the effective cwd and `agentDir`
- add a regression test that verifies the effective cwd/agentDir handoff

## Why
ATC subagent launches were failing immediately when user-scoped local package paths were configured in `~/.pi/agent/settings.json`. The subagent runner was creating the loader/settings manager without an explicit `agentDir`, which let pi package resolution fall through to `resolve(undefined, ...)` during package identity/deduping.

## Testing
- `npm test`
- `npm run typecheck`
- `npm run build`
- `npx vitest run test/agent-runner.test.ts`

### DIFF
--- a/src/agent-runner.ts
+++ b/src/agent-runner.ts
@@ -10,6 +10,7 @@ import {
   createAgentSession,
   DefaultResourceLoader,
   type ExtensionAPI,
+  getAgentDir,
   SessionManager,
   SettingsManager,
 } from "@mariozechner/pi-coding-agent";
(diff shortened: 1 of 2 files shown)


antikythera-technologies:
Closing this PR because the fork was pushed from the wrong account by mistake. Replacement PR: #30 from antikythera-agent-zero.

PR #39, opened 2026-04-28
Fix: subagent token count; Adds: Compaction / Context Window % to widget & agent responses, events

### LINKED ISSUE #38: Token count inside of the overlay is incorrect
The subagent overlay shows token totals 5–15× higher than actual usage. Root cause is upstream in  
  @mariozechner/pi-coding-agent:
                                                                                                     
  AgentSession.getSessionStats() computes                                                    
  total = Σ input + Σ output + Σ cacheRead + Σ cacheWrite
  across all assistant messages. But cacheRead per turn already represents the cumulative cached
  prefix re-read on that one API call, so summing it across N turns counts the prefix N times —      
  totals grow ~quadratically with turn count.                                                        
                                                                                                     
  pi-subagents reads this verbatim at:                                                               
  - src/ui/conversation-viewer.ts:108                                                        
  - src/ui/agent-widget.ts:310                                                                       
  - src/index.ts:127, :175, :347
                                                                                                     
  Workaround already available in upstream: getSessionStats() also returns contextUsage (computed via
   getContextUsage() → estimateContextTokens() → calculateContextTokens(lastAssistantUsage)), which  
  correctly uses only the latest assistant turn's usage and gives an accurate token figure. Switching
   the overlay to stats.contextUsage?.tokens (or, for billed-new-content: stats.tokens.input +       
  stats.tokens.output + stats.tokens.cacheWrite) avoids the inflation without waiting for upstream.

tintinweb:
Closes #38.                                                                                                                                                                      
                                         
  ## What changed                                                                                                                                                                  
   
  Replaced the upstream-derived token count with a lifetime accumulator on `AgentRecord` / `AgentActivity` (`{ input, output, cacheWrite }`), fed by a new `onAssistantUsage`      callback dispatched from `message_end` events in both `runAgent` and `resumeAgent`.
                                                                                                                                                                                   
  ## Why                                                    

  The previous count read from `getSessionStats().tokens.total`, which sums per-turn `cacheRead` across every assistant message — but each turn's `cacheRead` is the cumulative      cached prefix re-read on that one call, so summing across turns over-counts. And anything derived from `session.state.messages` resets at compaction, since upstream replaces the
   array. The new accumulator is independent of both, so the displayed total is monotone and accurate over a subagent's full lifetime.                                             
                                                            
  ## What's new

  - **Context-window utilization indicator** next to the token count: `(NN%)` sourced from `contextUsage.percent`, threshold-colored (<70% dim, 70–85% warning, ≥85% error).       
  - **Compaction count** in the same annotation slot: `(84% · ↻3)`.
  - **Parent-agent visibility** of all three signals — `get_subagent_result`, `steer_subagent`, and `task-notification` XML now return the subagent's current token count,           context-window %, and compaction count alongside the existing fields. Steering and result-fetch tool responses surface this *as the response*, so the main agent can react (e.g. 
  tighten its next steer when context is at 87%, or stop pushing more work into a subagent that's already compacted twice).                                                        
  - **`subagents:compacted` lifecycle event** with `{ id, type, description, reason, tokensBefore, compactionCount }`. Like the other `subagents:*` events, this is emitted via      `pi.events` and can be consumed by other pi extensions to react to subagent activity (e.g. dashboards, telemetry, automated escalation).                                         
  
  ## What's fixed                                                                                                                                                                  
                                                            
  - Token count no longer over-counts cached prefixes.
  - Token count no longer resets when a subagent compacts mid-run.
  - `subagents:completed`/`failed` event payload is now internally consistent: `input`, `output`, and `total` are all lifetime-accumulated together.



_________
Context: warning red
<img width="690" height="135" alt="image" src="https://github.com/user-attachments/assets/d5afb916-f794-4ce1-88a1-45cd99ecb577" />

_________
Agent response w/ context after compaction 110%->35%:
<img width="941" height="197" alt="image" src="https://github.com/user-attachments/assets/19c6d2a5-f537-426b-846c-66afb07a6320" />


tintinweb:
@copilot resolve the merge conflicts in this pull request

PR #42, opened 2026-04-30
Improvement: Move pi runtime packages from dependencies to peerDependencies

tintinweb:
xref https://github.com/tintinweb/pi-manage-todo-list/pull/1

> These packages are provided by the pi runtime at execution time. Listing them as `dependencies` causes unnecessary duplicate installs and, combined with the `"latest"` tag, widens the supply chain surface for no benefit.
>
> Moved to `peerDependencies` with `"*"` (any version the host provides) since the extension runs inside pi which already has these loaded.


tintinweb:
@copilot resolve the merge conflicts in this pull request

PR #45, opened 2026-05-03
Add new feature to force following the defined scoped models by user

y-almannaee:
### Summary

Adds `scopeModels` setting which is an opt-in validation that restricts subagent model choices to those listed in `enabledModels` from `~/.pi/agent/settings.json` which is selected via `/scoped-models`. 

This PR also has some UX improvements for the settings menu - `SettingsList` UI with numeric input support.

### Changes

**New module:** `src/enabled-models.ts` (+128 lines)
* `readEnabledModels()` - reads `enabledModels` array from global pi settings file (since it's not exposed in the API)
* `resolveEnabledModels()` - resolves exact provider/modelId patterns against the model registry (caches the list by `settings.json` mtime)

**Scope validation in** `src/index.ts` (+172/-39 lines)
* On agent spawn, resolved model is checked against `enabledModels`
* Agent's `params` -> hard error with allowed models listed
* Config/frontmatter -> warning to user + graceful fallback to parent/twin model to prevent interrupting the flow

**Settings:** `src/settings.ts` (+11 lines)
* Added `scopeModels?: boolean` to `SubagentsSettings` with sanitization and persistence

**Settings dialog rewrite:** `src/index.ts`
* Replaced `ctx.ui.select` menu with native `SettingsList` (`@mariozechner/pi-tui`)
* Numeric fields (`maxConcurrent`, `defaultMaxTurns`, `graceTurns`) support direct input via `ctx.ui.input` with validation
* Added `scopeModels` toggle to settings menu

**Tests**
* `test/enabled-models.test.ts` - 165 new lines, covers resolution, caching, edge cases (empty patterns, whitespace, corrupt JSON, deduplication, `getAvailable` filtering)
* `test/settings.test.ts` - 36 new lines for settings persistence

### DIFF
--- a/src/invocation-config.ts
+++ b/src/invocation-config.ts
@@ -24,8 +24,8 @@ export function resolveAgentInvocationConfig(
   isolation?: IsolationMode;
 } {
   return {
-    modelInput: agentConfig?.model ?? params.model,
-    modelFromParams: agentConfig?.model == null && params.model != null,
+    modelInput: params.model ?? agentConfig?.model,
+    modelFromParams: params.model != null,
     thinking: (agentConfig?.thinking ?? params.thinking) as ThinkingLevel | undefined,
     maxTurns: agentConfig?.maxTurns ?? params.max_turns,
     inheritContext: agentConfig?.inheritContext ?? params.inherit_context ?? false,
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,9 @@ Thumbs.db
 .claude
 .vscode
 .gitnexus
+.pi/subagents.json
+
+# Pi internal files
+progress.md
+AGENTS.md
+CLAUDE.md
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ Thumbs.db
 .pi/readcache
 .claude
 .vscode
+.gitnexus
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
       },
       "devDependencies": {
         "@biomejs/biome": "^2.3.5",
+        "@types/minimatch": "^5.1.0",
         "@types/node": "^25.5.0",
         "typescript": "^6.0.0",
         "vitest": "^4.0.18"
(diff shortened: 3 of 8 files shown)


tintinweb:
What is the rationale & usecase behind this?

y-almannaee:
The intent is to make subagent instantiation respect the same model scope the user already set in pi via `/scoped-models`.

Currently, the main agent can either manually select a model outside that scope or instantiate a subagent with frontmatter that specifies a model outside the scope. Users may: 

- scope pi to cheaper/local models, but an agent can still request Opus or another expensive model;
- scope pi to local providers only for sensitive workloads, but a subagent config can still route work to a cloud model;
- clone a repo which may include a model that is not enabled or intended for the user's environment.

Effectively this is for consistency: when I intentionally limit pi to a model set, I want spawned agents to also stay within that same set unless I disable the guard.

The feature is opt-in and off by default, so it does not change existing behavior as some people might rely on the agent configs with defined frontmatter, but enabling it also wont break agent configs, just warn the user and silently fall back. 

If the agent tries to define a subagent at runtime that falls outside the scope, then it will be told which models are in scope so it can course correct.

As an example, I had a subagent definition that defined a specific local model A while my the parent agent for that session used model B, but I only have enough compute for one local model. As a result, after the parent session instantiated that subagent they were fighting for execution and reloading each other in and out of memory, even though I only scoped Model B for that session.

src/invocation-config.ts
```diff
@@ -24,8 +24,8 @@ export function resolveAgentInvocationConfig(
   isolation?: IsolationMode;
 } {
   return {
-    modelInput: agentConfig?.model ?? params.model,
-    modelFromParams: agentConfig?.model == null && params.model != null,
+    modelInput: params.model ?? agentConfig?.model,
+    modelFromParams: params.model != null,
```
tintinweb, line 28: this changes the model selection precedence from "frontmatter is authorative" to "agent provided param is authorative"

tintinweb:
see #83 - fixes: reads enabledModels from cwd/global; frontmatter authorative; ...

PR #48, opened 2026-05-06
Fix: headless pi --print hangs/crashes with subagents

rylwin:
Fix two headless `pi --print` lifecycle issues around background subagents.

I found my pi --print processes would hang indefinitely when they used subagents. These fixes resolve the problem.

The commit messages include reproduction scripts that can be used to replicate the issue on master and show that the issue is resolved on this branch.

Happy to make any adjustments to the PR if helpful.

P.S. Thanks for this project -- it's exactly the subagent support I was looking for!


tintinweb:
🙏

PR #51, opened 2026-05-07
Set SessionName

forcepushdev:
Set the session name to be able to find out which agent is running in the session in an extension (e.g. agent-guardrails).


tintinweb:
Minor change:
> Manager-assigned id; suffixes session name to disambiguate parallel spawns (e.g. `Explore#a1b2c3d4`)

PR #52, opened 2026-05-08
feat(agent-tool): generate routing guidelines dynamically from agent descriptions

anh-chu:
The Agent tool currently has hardcoded routing hints in its description ("Use Explore for... Use Plan for..."). This PR makes the routing guidelines dynamic by reading all registered agent configs (default + custom) and generating hints from their descriptions.

**Change:** 1 file, +16/-3

**Before:** Static routing lines for Explore, Plan, general-purpose
**After:** Dynamic guidelines built from `getDefaultAgentNames()` and `getUserAgentNames()`, so custom agents (e.g. worker, reviewer) appear automatically in the LLM's routing hints.

Custom agents with descriptions get `- Use <name> for: <description>` lines. Built-ins without explicit descriptions keep their existing behavior via the default agent configs.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -601,6 +601,19 @@ export default function (pi: ExtensionAPI) {
 
   const typeListText = buildTypeListText();
 
+  /** Build routing guidelines dynamically from all available agent descriptions. */
+  const buildGuidelinesText = (): string => {
+    const allNames = [...getDefaultAgentNames(), ...getUserAgentNames()];
+    const routingLines = allNames.map((name) => {
+      const cfg = getAgentConfig(name);
+      const desc = cfg?.description ?? name;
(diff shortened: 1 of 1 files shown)


tintinweb:
Appears to duplicate `typeListText`

anh-chu:
true! I made this change a long while back so it may have been solving some specific routing problem I had. Something I noticed is that worker and general-purpose gets used a lot but Plan subagent, for example, almost never gets used. Any thoughts @tintinweb

tintinweb:
I have not observed this myself but while looking into your changes I was comparing our tool descriptions with claude code's and they deviate quite a bit. Might make sense to align this more w/o all the extra stuff they put there that might make no sense (and just waste tokens) 🙌

tintinweb:
I'll close this in favor of #65 which aligns with cc tool descriptions. Feel free to add to the discussion there.

PR #53, opened 2026-05-08
feat(agent): add chain mode with {previous} placeholder support

anh-chu:
Adds a `chain` parameter to the Agent tool that runs multiple agents sequentially, with each step receiving the previous step's output via a `{previous}` placeholder.

**Usage:**
```
Agent({
  chain: [
    { subagent_type: "Explore", prompt: "Find all auth-related files" },
    { subagent_type: "Plan", prompt: "Using: {previous}, design a JWT migration" },
    { subagent_type: "worker", prompt: "Implement: {previous}" },
  ]
})
```

**Changes:** 1 file, +228 lines in `src/index.ts`

- `chain` TypeBox parameter in Agent tool schema (Type.Optional Array, min 2 items)
- `executeChain()` function: sequential spawnAndWait, {previous} substitution, per-step error handling, abort support
- Summary output: "Chain completed in Xms (N steps)" with each step's result

When `chain` is provided, the single-agent path is skipped entirely. Chain steps inherit top-level params (model, thinking, isolation) as defaults.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -662,7 +662,8 @@ Guidelines:
 - Use model to specify a different model (as "provider/modelId", or fuzzy e.g. "haiku", "sonnet").
 - Use thinking to control extended thinking level.
 - Use inherit_context if the agent needs the parent conversation history.
-- Use isolation: "worktree" to run the agent in an isolated git worktree (safe parallel file modifications).${scheduleGuideline}`,
+- Use isolation: "worktree" to run the agent in an isolated git worktree (safe parallel file modifications).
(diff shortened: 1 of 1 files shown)


anh-chu:
Just to clarify, this means that users can prompt something like:
`explore, then plan -> review on plan, then implement`
Which would result in a streamlined E2E flow where each agent pipes their output to the next directly, avoiding polluting the main agent's context.

tintinweb:
autoclose: superseded. the sequential-pipeline use case is covered more generally by your #236 Workflow draft. reopen or file that one if you think this is the wrong call.

Issue #47, opened 2026-05-05
Subagents with `extensions: true` don't expose extension tools (MCP, etc.)


HyphaDev:
## Subagents with `extensions: true` don't expose extension tools (MCP, etc.)

Setting `extensions: true` in a subagent's frontmatter is documented as enabling extension access for that agent. In practice, extension-registered tools — including pi-mcp-adapter's `mcp` proxy — don't appear in the subagent's active tool set. Only the 7 built-in tools (`read, bash, edit, write, grep, find, ls`) are available.

## Reproduction

Create a custom agent at `~/.pi/agent/agents/mcp-enabled.md`:

```markdown
---
name: mcp-enabled
description: "Test"
extensions: true
inherit_context: false
---
You are a test subagent.
```

Then:

```
Agent({ subagent_type: "mcp-enabled", prompt: "List every tool you have access to, one per line." })
```

Returns:

```
read
bash
edit
write
grep
find
ls
```

No `mcp`, no other extension-registered tools. Expected: the `mcp` proxy plus any other extension-registered tools.

## Root cause

In `src/agent-runner.ts`, `runAgent` builds `sessionOpts` and passes them to `createAgentSession`:

```ts
const sessionOpts: Parameters<typeof createAgentSession>[0] = {
  // ...
  tools: toolNames,  // built-in names for the agent type
  resourceLoader: loader,
};
```

In pi-mono's `AgentSession`, the `tools:` option becomes `allowedToolNames`. The tool activation logic in `_refreshToolRegistry()` (agent-session.js) filters activation by that allowlist — even when `includeAllExtensionTools: true` is set in the same call:

```js
if (allowedToolNames) {
  for (const toolName of this._toolRegistry.keys()) {
    if (allowedToolNames.has(toolName)) {   // <-- extension tools never pass this
      nextActiveToolNames.push(toolName);
    }
  }
}
```

Extension tools ARE registered into `_toolRegistry` (because `loader.reload()` ran with `noExtensions: false`), but they never get activated because their names aren't in `allowedToolNames`.

The downstream `session.setActiveToolsByName(activeTools)` call in agent-runner tries to widen the active set, but `activeTools = session.getActiveToolNames().filter(...)` — which returns just the 7 builtins that were activated — can't re-add what wasn't there.

## Fix

When extensions are enabled, omit the `tools` allowlist:

```ts
// Before
tools: toolNames,

// After
tools: extensions === false ? toolNames : undefined,
```

[...]

PR #58, opened 2026-05-08
Feat: Allow agents to specify additional extension paths

forcepushdev:
Introduce `additional_extensions` field in agent frontmatter to load agent-specific extensions.

This enables developers to define extensions that are only active for a specific agent. The field supports CSV for multiple paths and `~` expansion for the user's home directory.

These agent-specific paths are passed to the resource loader, but are ignored when the agent is run in an isolated context.

### DIFF
--- a/src/agent-runner.ts
+++ b/src/agent-runner.ts
@@ -247,6 +247,9 @@ export async function runAgent(
     cwd: effectiveCwd,
     agentDir,
     noExtensions: extensions === false,
+    additionalExtensionPaths: extensions !== false
+      ? agentConfig?.additionalExtensionPaths
+      : undefined,
     noSkills,
     noPromptTemplates: true,
     noThemes: true,
--- a/src/custom-agents.ts
+++ b/src/custom-agents.ts
@@ -3,6 +3,7 @@
  */
 
 import { existsSync, readdirSync, readFileSync } from "node:fs";
+import { homedir } from "node:os";
 import { basename, join } from "node:path";
 import { getAgentDir, parseFrontmatter } from "@mariozechner/pi-coding-agent";
 import { BUILTIN_TOOL_NAMES } from "./agent-types.js";
@@ -70,6 +71,7 @@ function loadFromDir(dir: string, agents: Map<string, AgentConfig>, source: "pro
       isolation: fm.isolation === "worktree" ? "worktree" : undefined,
       enabled: fm.enabled !== false,  // default true; explicitly false disables
       source,
+      additionalExtensionPaths: parseExtensionPaths(fm.additional_extensions),
     });
   }
 }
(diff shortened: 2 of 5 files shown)


anh-chu:
hey this sounds interesting! Do you mind sharing what extensions are you using for subagents only? @forcepushdev

forcepushdev:
@anh-chu I wrote some agent guardrails that help me to allow/deny more specific then "read/write/...".
My Planner-Agent is only allowed to write to PLAN.md, the reviewer to my review.json etc.

This is something I am not able to force with the currently available options in the md file.

tintinweb:
We should probably rework the semantics around `extensions: [true,false,[list]]` and `tools:` #47 #64 together with allowing to specify additional extension paths. I can see how this is useful. Trying to figure out if the path should go into `extensions: true, false, [name, path, path, path]` or if we really want an `additional_extensions: [path]` field 🤔 

- `additional_extensions` is a new field (cons) but allows to easily add extensions on top of default settings. (only exposed to frontmatter; not agent param)
- `extensions:` is more compact, implicit but if someone wants to have default extensions on (`extensions: true`) + these 3 extra extensions then they would need to specify all of them in the list. not sure if I like that.  this option is less flexible in case someone does not want default extensions + some extensions specified by path.

I tend to an extra field right now. but need to rework `extensions` with `tools` anyway. The main question there is: extensions register tools, so the `tools` field is like a master switch. what happens if someone specifies `extensions: [ext1, ext2]`. would we assume all tools from ext1,ext2 allowed or should this still be gated through `tools`.

lmk if anyone has thoughts on this. I tend to `tools` as the master gate, `extensions:` could gate loaded extensions by name (default all on) and `additional_extensions` could just be an easy way to have the subagent load more.

forcepushdev:
@tintinweb I can find good things in both. I decided to use a new field because I think true, false or [string array] is kinda strange.
I would prefer to keep extensions as it is, indicating to use the parents extensions (maybe rename it to smt. like `inherit_extensions`) and have a new field for additional ones.
Otherwise you have to add all extensions you want to inherit into the extension field as soon as you want to add extra ones.

I would tend to your suggestion
'lmk if anyone has thoughts on this. I tend to tools as the master gate, extensions: could gate loaded extensions by name (default all on) and additional_extensions could just be an easy way to have the subagent load more.'
too. This seems to be the most consistent approach.

anhchu-katalon:
Yes. I also like your model @tintinweb . My original use case was around extensions that do not load tools, but manipulate some system calls required for consistent behavior across main agent and subagents. This sounds like a good solution.

tintinweb:
I have a working draft ready that allows flexible selection and loading of extensions and tools. I think I will skip the extra `addtional_extensions` in favor of `extension: <path>, name` (this controls what extension is loaded) and `tools:` controls what agent or extension tools are offered. both support wildcard `*` to select all tools - and I'll probably have an `all|true` alias. I think that this makes more sense than more fields with a slight cost of semantic overload but it allows more flexibility and is more descriptive. I'll try to land this eof week. just got back from some travels :)

tintinweb:
Thanks for the proposal.
  
  **How #74 addresses it:** path entries (`~/foo.ts`, `/abs/foo.ts`, `./local.ts`) sit directly inside the existing `extensions: [...]` field alongside name entries and `*` — no separate `additional_extensions` field, same CSV / `~` / isolated-ignored behavior. Details in #74 / CHANGELOG.
  
  Closing in favor of #74 on merge — please test the path form once released and reopen if your use case isn't covered.

PR #59, opened 2026-05-08
fix(rpc): resolve string model in subagents:rpc:spawn handler

### LINKED ISSUE #60: TaskExecute(model: "…") fails with "No API key found for undefined" — RPC spawn doesn't resolve string model
## Symptom

Spawning a subagent through `@tintinweb/pi-tasks@>=0.4.3`'s `TaskExecute` with a `model` override fails immediately:

```
Error: No API key found for undefined.
```

The provider is `undefined` because pi later reads `.provider` on a string instead of a `Model` object.

## Reproduction

- `TaskExecute({ model: "openai-codex/gpt-5.5", … })` → fails as above
- `TaskExecute` with no model (inherits parent) → works
- `Agent` tool spawn with `model: ctx.model` (object) → works
- Auth itself is fine for the model: `pi auth add openai-codex …` is set up; same models work when invoked with object form

Reproduced against:
- `@tintinweb/pi-subagents@0.7.1`
- `@tintinweb/pi-tasks@0.5.0`

## Source evidence

`pi-tasks` forwards the string from tool params directly into spawn options (`pi-tasks/src/index.ts`):

```ts
...(params.model ? { model: params.model } : {}),
```

… introduced in [tintinweb/pi-tasks#7](https://github.com/tintinweb/pi-tasks/pull/7), included in `0.4.3` and `0.5.0`.

`pi-subagents` `SpawnOptions` expects:

```ts
model?: Model<any>;
```

`runAgent()` uses `options.model` directly:

```ts
// src/agent-runner.ts:260
const model = options.model ?? resolveDefaultModel(...);
```

So a string bypasses model resolution and pi later trips on `.provider` access.

## Why fix in `pi-subagents`

`subagents:rpc:spawn` is a public cross-extension boundary. RPC callers naturally send serializable values (a string `"provider/modelId"`) — they shouldn't need internal pi `Model` objects. The scheduler path inside this same package already calls `resolveModel(job.model, ctx.modelRegistry)`; the RPC handler should do the same.

## Proposed fix

Normalize `options.model` inside the `subagents:rpc:spawn` handler in `src/cross-extension-rpc.ts` before `manager.spawn()`:

- if `options.model` is a string, call `resolveModel(options.model, ctx.modelRegistry)`
- if resolution fails, surface the human-readable `Model not found` error (better signal than silent fallback)
- if `ctx.modelRegistry` is missing while a string was supplied, error explicitly (configuration bug)

Object models continue to pass through unchanged.

## PR

Patch + tests in #59.

any-victor:
Closes #60.

Fixes a regression for cross-extension callers (e.g. `@tintinweb/pi-tasks@>=0.4.3`'s `TaskExecute`) that forward a serializable `model: "provider/modelId"` string into the RPC. `runAgent()` expects a `Model` object, so the spawned agent fails with:

```
Error: No API key found for undefined.
```

because pi reads `.provider` on the string.

The scheduler path inside this same package already uses `resolveModel(job.model, ctx.modelRegistry)` to handle the same shape; this PR applies the same pattern in the `subagents:rpc:spawn` handler.

## What changes

- `src/cross-extension-rpc.ts`: when `options.model` is a string, call `resolveModel(options.model, ctx.modelRegistry)`. Object models pass through unchanged. If the registry isn't on `ctx`, the handler errors clearly. If the string can't be resolved, the human-readable `Model not found: "…"` message is surfaced instead of a confusing 'No API key for undefined'.
- `test/cross-extension-rpc.test.ts`: 4 new cases under 'spawn RPC model override' covering each path (string → Model, object passthrough, unresolved string, missing registry).
- `package.json`: adds `@mariozechner/pi-agent-core` to `peerDependencies`. `src/types.ts` imports `ThinkingLevel` from it but the package was never declared, so `tsc --noEmit` on a fresh clone always failed. Also added as `devDependency` so local dev resolves it without a manual install.
- `pnpm-workspace.yaml`: `allowBuilds` for `@google/genai`, `koffi`, `protobufjs` so installs don't gate on interactive approval.

## Test plan

```
pnpm install --no-frozen-lockfile
pnpm run typecheck    # clean
pnpm run lint         # clean
pnpm run test         # 366 / 366 passing across 22 files
```

## Background

`@tintinweb/pi-tasks` forwards `params.model` directly into spawn options:

```ts
// pi-tasks/src/index.ts
...(params.model ? { model: params.model } : {}),
```

So `TaskExecute({ model: "openai-codex/gpt-5.5" })` reaches `runAgent()` with a string, fails the auth lookup. Same models work when invoked with `model: ctx.model` object.

The fix resolves at the RPC boundary, the natural place for serializable cross-extension input.


tintinweb:
hey @any-victor, thanks for the PR. can you remove lockfiles and the dependency updates from the PRs? I can't really take them for security reasons. Thanks 🙌

any-victor:
Thanks for the heads-up. I removed the lockfiles and dependency changes from both PRs, so they should now only include the source and test changes.

PR #64, opened 2026-05-11
Fix to allow extension tools with isolated: false

rafleon:
This fixes this issue:

https://github.com/tintinweb/pi-subagents/issues/47

### DIFF
--- a/src/agent-runner.ts
+++ b/src/agent-runner.ts
@@ -271,7 +271,10 @@ export async function runAgent(
     settingsManager: SettingsManager.create(effectiveCwd, agentDir),
     modelRegistry: ctx.modelRegistry,
     model,
-    tools: toolNames,
+    // When extensions are enabled, omit the tools allowlist so extension-registered
+    // tools (MCP, LSP, etc.) in the registry can be activated. Otherwise they would
+    // be filtered out by allowedToolNames in pi-mono's AgentSession constructor.
+    tools: extensions === false ? toolNames : undefined,
     resourceLoader: loader,
   };
   if (thinkingLevel) {


tintinweb:
Thanks for the PR @rafleon. This needs a little more investigation, see https://github.com/tintinweb/pi-subagents/pull/58#issuecomment-4425097156. I'll keep this open until I have clarity on the semantics 🙌

nekwebdev:
I reproduced this with custom agent frontmatter like:

 ```yaml
 tools: read, grep, find, ls
 extensions: web, fetch
 ```

This PR fixes the main root cause I saw: `createAgentSession({ tools: toolNames })` acts as a global allowlist, so extension tools never enter the child session registry. That means tools like `web_search` / `fetch_content` cannot be recovered later by the `extensions:` allowlist.

In context of #58, I think this needs to be resolved first (or at least together with #58). #58 adds `additional_extensions`, but those agent-specific extensions would hit the same issue: if `tools` is passed as the session-wide allowlist, tools registered by those additional extensions can still be excluded before the agent can use them.

So I see #64 as a prerequisite/baseline for #58’s behavior to be useful for tool-providing extensions. #72 also looks complementary: it handles tools that appear during/after `bindExtensions`, then reapplies the allowlist/denylist.

Ideal sequencing/shape seems:

- land this PR’s fix so `tools:` does not accidentally prevent extension tools from being registered
- add #72’s post-`bindExtensions` re-filter so late extension tools are also gated correctly
- then #58’s `additional_extensions` can build on clear `tools` / `extensions` semantics

Without this PR, #58 may load extra extension paths but still not expose their tools. Without #72, tools registered only after extension binding may not be filtered consistently.

Just more noise to help you find that clarity :) and thanks for your work on this very helpful project!

tintinweb:
Thanks for taking a swing at #47.

  **How #74 fixes it:** master allowlist built at construction (different approach than this PR's post-bind re-filter). Details in #74 / CHANGELOG.
  
  Closing in favor of that on merge — please test once a new release ships and reopen if anything's still broken for you.

PR #72, opened 2026-05-12
fix(agent-runner): re-filter active tools after bindExtensions so extension tools land in child

gotgenes:
## Summary

`runAgent` filters the child session's active tool names **before** calling `session.bindExtensions(...)`. Extension-registered tools are added to the session during `bindExtensions`, so the filter pass never sees them — making the `extensions: string[]` allowlist branch functionally dead for extension tools and letting denylisted extension tools through under `extensions: true` + `disallowedTools`.

Related: #47

## Fix

Extract the filter into a `filterActiveTools` helper and call it both before `bindExtensions` (current behavior, filters built-in tools) and after (new, filters extension-registered tools). The post-bind pass uses identical rules against the post-bind active set.

## Tests

Adds 7 new tests in `test/agent-runner-extension-tools.test.ts` covering:

1. `setActiveToolsByName` is called both before and after `bindExtensions`
2. Post-bind re-filter includes extension tool when `extensions: true`
3. Post-bind re-filter respects `extensions: string[]` allowlist
4. Post-bind re-filter respects `disallowedTools` denylist for extension tools
5. Post-bind re-filter excludes our own tools (`EXCLUDED_TOOL_NAMES`) even when `extensions: true`
6. `extensions: false` still applies `disallowedTools` to built-in tools (post-bind re-filter)
7. `extensions: false` with no `disallowedTools` skips the filter (no `setActiveToolsByName` call)

Negative-control validated: the new tests fail against the unpatched code.

## Verification

`npm run typecheck && npm test` passes (379 upstream + 7 new = 386 tests).

## Relation to #64

PR #64 addresses the same root cause (#47) with a different approach: passing `tools: undefined` when extensions are enabled so all registered tools activate by default. This PR takes an alternative approach — keeping the existing `tools` allowlist but re-running the filter after `bindExtensions` so extension tools are gated by the same rules (allowlist, denylist, EXCLUDED_TOOL_NAMES) as built-in tools.

Either approach fixes the bug. This one preserves the invariant that the session's active set is always explicitly filtered rather than briefly containing everything.

## Reference implementation

This change has been running in production via [`@gotgenes/pi-subagents@0.8.0`](https://github.com/gotgenes/pi-subagents) (a friendly fork). See [Tiny-IG-Software/repone#445](https://github.com/Tiny-IG-Software/repone/issues/445) for context.

### DIFF
--- a/src/agent-runner.ts
+++ b/src/agent-runner.ts
@@ -26,6 +26,42 @@ import type { SubagentType, ThinkingLevel } from "./types.js";
 /** Names of tools registered by this extension that subagents must NOT inherit. */
 const EXCLUDED_TOOL_NAMES = ["Agent", "get_subagent_result", "steer_subagent"];
 
+/**
+ * Filter the session's active tool names according to extension/denylist rules.
+ *
+ * Run twice — once before `bindExtensions` (filters built-in tools) and once after
+ * (filters extension-registered tools, which only join the active set during
+ * `bindExtensions`). Extracting this keeps the two callsites consistent and makes
+ * the post-bind re-filter trivial.
+ *
+ * @param activeTools  Names currently active on the session.
+ * @param toolNames    The built-in tool name allowlist for this agent type.
+ * @param extensions   Agent config `extensions` field: false | true | string[] (allowlist).
+ * @param disallowedSet  Optional denylist from agent config.
+ */
+function filterActiveTools(
+  activeTools: string[],
+  toolNames: string[],
+  extensions: boolean | string[],
+  disallowedSet: Set<string> | undefined,
+): string[] {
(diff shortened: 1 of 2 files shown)


tintinweb:
Thanks for the diagnosis — pre/post-bind split was exactly the right framing.
  
  **How #74 fixes it (differently):** instead of running the filter twice (pre + post `bindExtensions`), all gating moves to session construction via the master allowlist, so `setActiveToolsByName` is no longer called at all. Your 7 scenarios are validated equivalently in the new tests. Details in #74 / CHANGELOG.
   
  Closing in favor of #74 on merge — please give it a try once released and reopen if anything's not behaving as expected.

PR #73, opened 2026-05-12
feat(prompts): inject <active_agent name="..."/> tag for permission resolution

gotgenes:
## Summary

Prepends `<active_agent name="${config.name}"/>` to every assembled child system prompt (both `replace` and `append` modes) so downstream extensions can resolve per-agent policy inside the child by parsing the system prompt.

## Motivation

Downstream permission/policy extensions need to know which agent is currently active inside a child session to apply per-agent policy (e.g. per-agent `permission:` frontmatter). The child session has no API surface exposing this directly — the only stable signal that survives session-creation is the system prompt itself.

This change provides a small, parseable marker (`<active_agent name="..."/>`) without coupling `pi-subagents` to any specific downstream extension. The tag is:
- **Short** — one line, prepended before the env block
- **Harmless if ignored** — models treat unknown XML-style tags as structural hints
- **Unambiguous to parse** — regex: `<active_agent\s+name=["\x27]([^"\x27]+)["\x27][^>]*>`

## Change

```diff
 export function buildAgentPrompt(...): string {
+  const activeAgentTag = `<active_agent name="${config.name}"/>\n\n`;
+
   const envBlock = ...;
   ...
   // append mode
-  return envBlock + "\n\n<inherited_system_prompt>\n" + ...;
+  return activeAgentTag + envBlock + "\n\n<inherited_system_prompt>\n" + ...;
   ...
   // replace mode
-  return replaceHeader + "\n\n" + config.systemPrompt + extrasSuffix;
+  return activeAgentTag + replaceHeader + "\n\n" + config.systemPrompt + extrasSuffix;
 }
```

## Tests

Adds 4 new tests in `test/prompts.test.ts`:

1. Tag is present at start of prompt in `replace` mode
2. Tag is present at start of prompt in `append` mode
3. Tag uses agent name verbatim (no escaping or normalization)
4. Tag appears before the env block in both modes

Negative-control validated: the new tests fail against the unpatched code.

## Verification

`npm run typecheck && npm test` passes (379 upstream + 4 new = 383 tests).

## Reference implementation

This change has been running in production via [`@gotgenes/pi-subagents@0.8.0`](https://github.com/gotgenes/pi-subagents) (a friendly fork). The patch is at https://github.com/gotgenes/pi-subagents/commit/83c225b. See [Tiny-IG-Software/repone#445](https://github.com/Tiny-IG-Software/repone/issues/445) for context.

### DIFF
--- a/src/prompts.ts
+++ b/src/prompts.ts
@@ -19,6 +19,10 @@ export interface PromptExtras {
  * - "append" mode: env header + parent system prompt + sub-agent context + config.systemPrompt
  * - "append" with empty systemPrompt: pure parent clone
  *
+ * Both modes prepend an `<active_agent name="${config.name}"/>` tag so downstream
+ * extensions (e.g. permission/policy systems) can resolve per-agent policy
+ * inside the child session by parsing the system prompt.
+ *
  * @param parentSystemPrompt  The parent agent's effective system prompt (for append mode).
  * @param extras  Optional extra sections to inject (memory, preloaded skills).
  */
@@ -29,6 +33,8 @@ export function buildAgentPrompt(
   parentSystemPrompt?: string,
   extras?: PromptExtras,
 ): string {
+  const activeAgentTag = `<active_agent name="${config.name}"/>\n\n`;
+
   const envBlock = `# Environment
 Working directory: ${cwd}
 ${env.isGitRepo ? `Git repository: yes\nBranch: ${env.branch}` : "Not a git repository"}
@@ -66,7 +72,7 @@ You are operating as a sub-agent invoked to handle a specific task.
       ? `\n\n<agent_instructions>\n${config.systemPrompt}\n</agent_instructions>`
       : "";
 
-    return envBlock + "\n\n<inherited_system_prompt>\n" + identity + "\n</inherited_system_prompt>\n\n" + bridge + customSection + extrasSuffix;
+    return activeAgentTag + envBlock + "\n\n<inherited_system_prompt>\n" + identity + "\n</inherited_system_prompt>\n\n" + bridge + customSection + extrasSuffix;
   }
 
   // "replace" mode — env header + the config's full system prompt
@@ -75,7 +81,7 @@ You have been invoked to handle a specific task autonomously.
 
 ${envBlock}`;
 
-  return replaceHeader + "\n\n" + config.systemPrompt + extrasSuffix;
+  return activeAgentTag + replaceHeader + "\n\n" + config.systemPrompt + extrasSuffix;
 }
 
 /** Fallback base prompt when parent system prompt is unavailable in append mode. */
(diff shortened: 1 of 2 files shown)


jeffutter:
👋 Hey, I’ve been looking at optimizing some of my prompt processing times with local LLMs with llama.cpp in pi. I use this subagent plugin (well @gotgenes ‘s fork) and I noticed that when every subagent kicks off my LLM loses any prompt cache. Ideally, especially for “append” subagents llama.cpp should be able to reuse a bunch of the prompt processing cache. Prepending the system prompt with anything breaks that.

I’m not sure I understand the full context of the purpose of this change but I was wondering if you had considered prompt caching and if there is another way to achieve the same goal without prepending the system prompt?

gotgenes:
I'm unfamiliar with "append" subagents. Can you tell me more or point me to some documentation? I'm trying to better understand your use case.

I recognize this busts the cache compared to the parent's prompt but I guess I would not expect a fresh subagent to benefit from prompt caching at its inception in the first place. But obviously if it _could_, that's desirable.

jeffutter:
The append behavior is briefly covered here: https://github.com/tintinweb/pi-subagents#frontmatter-fields

> prompt_mode 
>     replace: body is the full system prompt (no AGENTS.md / CLAUDE.md inheritance). 
>     append: body appended to parent's prompt (agent acts as a "parent twin" — inherits parent's AGENTS.md / CLAUDE.md)

I think this is indicating that `prompt_mode` append (which many of the default subagents except maybe research, and most of my custom ones are) should share the system prompt and AGENTS.md with the main agent.

For some more concrete numbers, I grabbed the system prompt from the beginning of a session and the system prompt from the invocation of a `scout` agent. 

They share `8,333` tokens.


The layout looks like this:

```
<active_agent name="scout"/>

# Environment
Working directory: /home/jeffutter/src/my-project
Git repository: yes
Branch: main
Platform: linux

<inherited_system_prompt>

.... 8333 shared tokens - system prompt and AGENTS.md list of skills ....

</inherited_system_prompt>

<sub_agent_context>
You are operating as a sub-agent invoked to handle a specific task.
- Use the read tool instead of cat/head/tail
- Use the edit tool instead of sed/awk
- Use the write tool instead of echo/heredoc
- Use the find tool instead of bash find/ls for file search
- Use the grep tool instead of bash grep/rg for content search
- Make independent tool calls in parallel
- Use absolute file paths
- Do not use emojis
- Be concise but complete
</sub_agent_context>

<agent_instructions>
You are the **scout agent**. Your job is simple: mark the task as picked up and return its full text — nothing else.

1. Mark the task as Needs Plan:
   `mcp({ tool: "backlog_task_edit", args: '{"id": "<task-id>", "status": "Needs Plan"}' })`

2. Fetch the full task text:
   `mcp({ tool: "backlog_task_view", args: '{"id": "<task-id>"}' })`

Output only the raw task text. Do not summarise, annotate, or add commentary.
</agent_instructions>
Current date: 2026-05-22
Current working directory: /home/jeffutter/src/my-project
```

The preamble loses the entire cache on that 8,333 tokens which, on my sad hardware takes something like 40 seconds to process 🤣

PR #79, opened 2026-05-20
fix(agent): emit lifecycle events for foreground RPC agents

micuintus:
When a subagent is spawned via cross-extension RPC with `isBackground: false`, the `subagents:completed` event is never emitted.

`agent-manager.ts` only calls `onComplete` (which emits the event) for background agents. Foreground agents are expected to have their caller await the promise directly, but an RPC caller has no promise to await. Cross-extension consumers like `pi-dacmicu/evolve` rely on the event to detect completion.

Changes:
- Emit `onComplete` for ALL agents (background and foreground)
- Add `isBackground` field to `AgentRecord` to preserve the spawn flag
- Gate notification / batching on `isBackground` in `index.ts`; foreground agents still update the widget but do not receive follow-up nudges
- Keep `runningBackground--` guarded so the concurrency counter stays accurate.

### DIFF
--- a/src/agent-manager.ts
+++ b/src/agent-manager.ts
@@ -125,8 +125,8 @@ export class AgentManager {
       startedAt: Date.now(),
       abortController,
       lifetimeUsage: { input: 0, output: 0, cacheWrite: 0 },
-      compactionCount: 0,
       invocation: options.invocation,
+      isBackground: options.isBackground ?? false,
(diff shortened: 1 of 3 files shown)


micuintus:
Closing: branched from wrong parent, picked up unrelated widget fix. Will reopen from clean branch.

PR #85, opened 2026-05-29
feat: markdown result preview + failure-body fallback + uncapped success output

cad0p:
Brings markdown rendering + failure-body fallback to subagent completion notifications. Full design doc and 4-phase release-grade methodology audit trail (4 implementation review rounds + E2E with live pi + 4 bloat-reduction review rounds) in cad0p/tintinweb-pi-subagents#1 — 47 commits squashed to one for review here.

## What changes

- `subagent-notification` body now renders via `new Markdown(body, 2, 0, getMarkdownTheme())` (using `@earendil-works/pi-tui` + `@earendil-works/pi-coding-agent` exports already in use). Headings, bullets, tables, code fences, syntax highlighting all theme-styled.
- **`resultPreviewMode: "plain" | "markdown"`** (default `"markdown"`). Plain mode is byte-equivalent to the current renderer; pinned by `test/renderer-regression.test.ts` snapshots.
- **`resultPreviewExpanded: boolean`** (default `true`). When true, full body inline; `ctrl+O` is a no-op. When false, collapsed to first 10 lines + `… (N more lines, ctrl+O to expand)` hint (matches `read` / `write` tool precedent at `core/tools/read.js:112`, `write.js:109`).
- **`failurePreviewMaxChars: number`** (default 65536). Cap applies only on `error` / `stopped` status; success / aborted / steered are uncapped. Removes the 500-char (individual) / 300-char (group) caps on success-path output, which forced `get_subagent_result` round-trips for any non-trivial result.
- **Failure-body fallback** `record.result ?? record.error ?? ""` in both `formatTaskNotification` and `buildNotificationDetails`. Failed agents previously rendered `"No output."` because `record.result` is undefined on `error` / `stopped` paths (only set on success in `agent-manager.ts:226, :358`); the actual error text was visible only in the XML `<status>` field.

## Backward compatibility

Default behavior changes (markdown rendering + uncapped success output) but the upstream behavior is fully reachable via `resultPreviewMode: "plain"` + `resultPreviewExpanded: false`. Snapshot test in `test/renderer-regression.test.ts` locks byte-equivalence (modulo timestamps).

## Tests

- 500 tests pass (`npm test`)
- Typecheck/lint/build clean
- `stripInternal: true` added to `tsconfig.json` so `@internal` testability seams (e.g. `subagentNotificationRenderer`, `buildResultPreview`) don't leak into `dist/index.d.ts`

## Settings UI

Three new entries in `/agents → Settings`. Persist to existing `.pi/subagents.json` via the same global/project precedence as the existing settings.

Happy to split or scope this differently if you'd prefer — the design has no hard coupling between the markdown rendering, the cap removal, and the failure-body fallback.


ecbrodie:
@cad0p are you still pursing merging this PR? If not, would you be cool if I took it over? I'd love to get this shipped, especially the part where `renderResult` is used to render the subagent result body into the Pi session.

cad0p:
Cool with you taking over as long as it gets my fork closer to upstream! https://github.com/cad0p/pi-subagents-tintinweb

I also did some improvements regarding avoiding double notification when the agent is actively calling `get_subagent_result` and the subagent is done. 

Also, the completion notification gets queued as steering instead of follow up, so the completion notification gets delivered asap. 

I also added a checkpoint tool for the parent to know how far the subagent is as I noticed the parent didn’t have enough info about the subagent.

PR #87, opened 2026-05-30
feat(tools): add promptSnippet and promptGuidelines to registered tools

that-yolanda:
## Problem

The three tools registered by pi-subagents (`Agent`, `get_subagent_result`, `steer_subagent`) do not provide `promptSnippet` and `promptGuidelines` parameters when calling `defineTool` / `pi.registerTool`.

In pi-mono's system prompt assembly logic (`system-prompt.ts`), these two fields serve critical roles:
- **`promptSnippet`**: A one-line description that appears in the system prompt's "Available tools" section. Tools without this field are omitted from that section entirely.
- **`promptGuidelines`**: Guideline bullets appended to the system prompt's "Guidelines" section.

**Impact**: Without these fields, the three tools are completely absent from the default system prompt. For models that strongly follow system prompt instructions (e.g., Claude), this may cause them to overlook tool schema parameter descriptions, leading to suboptimal tool usage.

## Changes

Only one file modified: `src/index.ts`

### 1. Agent tool — added `promptSnippet` + `promptGuidelines`

```typescript
promptSnippet: "Launch autonomous sub-agents for complex multi-step tasks",
promptGuidelines: [
  "Use Agent with specialized agents when the task matches an agent type's description...",
  "For broad codebase exploration or research, spawn Agent with an appropriate subagent_type...",
  "When an agent runs in the background, you will be notified on completion — do not poll...",
  "Trust but verify: an agent's summary describes intent, not outcome...",
],
```

The guidelines are aligned with Claude Code's system prompt conventions (specifically the "Session-specific guidance" section), ensuring compatibility with models like Claude while remaining generic enough for any LLM.

### 2. get_subagent_result — added `promptSnippet` only

Simple utility tool — the schema description is sufficient; a snippet ensures it appears in "Available tools".

### 3. steer_subagent — added `promptSnippet` only

Same rationale as above.

## Verification

- ✅ TypeScript compilation: `npm run typecheck` passes
- ✅ All 25 test files, 458 test cases pass: `npm test`
- ✅ Diff is minimal: 1 file, +9 lines

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -695,6 +695,13 @@ Provide clear, detailed prompts so the agent can work autonomously. Brief it lik
 Terse command-style prompts produce shallow, generic work.
 
 **Never delegate understanding.** Don't write "based on your findings, fix the bug" or "based on the research, implement it." Those phrases push synthesis onto the agent instead of doing it yourself. Write prompts that prove you understood: include file paths, line numbers, what specifically to change.`,
+    promptSnippet: "Launch autonomous sub-agents for complex multi-step tasks",
+    promptGuidelines: [
+      "Use Agent with specialized agents when the task matches an agent type's description. Subagents are valuable for parallelizing independent queries or for protecting the main context window from excessive results, but should not be used excessively when not needed. Importantly, avoid duplicating work that subagents are already doing — if you delegate research to a subagent, do not also perform the same searches yourself.",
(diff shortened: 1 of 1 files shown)


tintinweb:
Thanks for picking this up. I traced how pi assembles the prompt and this matches how the built-in tools use these
  fields, so good to merge.

  One optional note: the tools aren't actually invisible without these fields — their full description and params still
  reach the model via the tool schema. These fields mainly add a reminder in the system prompt prose, which helps
  prompt-following models like Claude.

PR #98, opened 2026-06-10
feat: exclude_extensions frontmatter — extension denylist for subagents.

### LINKED ISSUE #94: Feature request: Add support for excluding specific extensions from subagents
## Feature request

It would be useful to support an extension exclude/denylist option for subagents.

### Current behavior

Agent frontmatter currently allows controlling extensions with options like:

```yaml
extensions: true
extensions: false
extensions: [mcp, other-extension]
extensions: ["*", "/abs/path/to/extension.ts"]
```

This works well when I want either all extensions, no extensions, or a manually maintained allowlist.

However, there does not seem to be a way to say:

```yaml
extensions: true
exclude_extensions: [pi-notify]
```

or:

```yaml
extensions: ["*", "!pi-notify"]
```

### Use case

I use `pi-subagents` together with `pi-notify`.

I want my main Pi agent/session to show a desktop notification when it finishes. However, I do not want every subagent to trigger its own completion notification.

At the same time, I still want subagents to use most of my other installed extensions. Using:

```yaml
extensions: false
```

is too restrictive because it disables all extensions for the subagent.

Using an explicit allowlist like:

```yaml
extensions: [mcp, other-extension, another-extension]
```

works, but it is inconvenient to maintain because every subagent needs to manually list every extension except the notification extension.

### Proposed solution

Add support for an extension denylist/exclude option, for example:

```yaml
---
description: Research agent
extensions: true
exclude_extensions: [pi-notify]
tools: "*"
---
```

This would mean:

* Load the normal/default extension set.
* Remove the listed extensions from that set.
* Continue allowing all other extensions.

### Alternative syntax

Another possible syntax could be glob-style negation:

```yaml
extensions: ["*", "!pi-notify"]
```

But a separate field such as `exclude_extensions` may be clearer and easier to validate.

### Why this matters

Some extensions are useful in subagents, while others are only useful in the main agent/session.

Notification extensions are a good example: they are helpful when the main agent finishes, but noisy when multiple background subagents complete independently.

An exclude option would make it easier to keep subagents powerful while preventing specific side-effect extensions from running inside subagent sessions.

### Expected behavior

Given:

```yaml
---
extensions: true
exclude_extensions: [pi-notify]
---
```

The subagent should load all default extensions except `pi-notify`.

The main agent/session should still be able to use `pi-notify` normally.

tintinweb:
## Summary

  Adds an `exclude_extensions:` agent-frontmatter field — an extension **denylist** applied after the `extensions:`
  include set. Closes #94 (thanks @ramhaidar for the report).

  The key use case: run subagents with all extensions *except* a noisy one, without hand-maintaining an allowlist:

  ```yaml
  ---
  description: Research agent
  extensions: true
  exclude_extensions: pi-notify
  ---
  ```

  ## Design

  - **Exclude wins** — over `extensions: true`, `"*"`, allowlists, and `tools: ext:` selectors (an excluded extension
  never loads, so an `ext:` reference becomes the existing orphan warning).
  - **Plain canonical names only** (case-insensitive). No paths, no `*`, and deliberately **no `["*", "!name"]` negation
  syntax** — an unquoted `!name` is a YAML tag and silently mis-parses.
  - **Warn-not-abort**, consistent with existing `extensions:` mismatch handling (`extension-error:…` via
  `onToolActivity`):
    - exclude matching no discovered extension → typo warning (checked against the *pre-filter* discovered set)
    - `extensions: false` + exclude → "has no effect" warning
    - name in both `extensions:` and `exclude_extensions:` → accurate "exclude wins" message (previously this shape
  would have warned misleadingly "was not loaded")
  - **Not a sandbox** (documented): excluded extensions' factory code still executes once during `loader.reload()` —
  exclusion suppresses handler binding and tool registration, not load-time side effects.
  - `isolated: true` nulls excludes along with everything else (intentional override, no warnings).

  ## Implementation

  - `src/types.ts`, `src/agent-types.ts`, `src/custom-agents.ts` — optional `excludeExtensions` threaded frontmatter →
  `AgentConfig` → `getConfig()`; parsed with the existing `csvListOptional` (CSV or YAML list; `none`/omitted → unset)
  - `src/agent-runner.ts` — the core: `extensionsOverride` is now also installed when load-all + excludes exist; single
  filter predicate (exclude first, then include set); the override callback captures the pre-filter discovered set for
  the typo warning
  - `src/index.ts` — `/agents → Eject` serializes the field back to frontmatter
  - README field table + scoping rules (incl. the not-a-sandbox caveat), CHANGELOG

  ## Purely additive — verified

  The no-exclude path is provably identical to before (the new condition and filter reduce to the old ones when no
  excludes exist). Confirmed by running the **unmodified pre-change test suite against the new source: 561 passed, 0
  failures**. Only user-visible change without the feature: one warning hint string, `(add it to extensions:)` → `(check
  extensions:/exclude_extensions:)`.

  ## Tests (14 new)

  - 3 frontmatter-parse tests (CSV, YAML list, `none`/omitted)
  - 8 runner tests: `extensions: true` + exclude, wildcard no longer short-circuits, allowlist subtract + "in both"
  warning, typo warning, orphan warning, `isolated:` suppression, `ext:` selector can't resurrect an excluded extension,
  case-insensitivity
  - 2 auto-discovered e2e templates through real pi-mono: `exclude-beta`, `exclude-beats-ext-selector`

  Full suite: 572 passed / 4 skipped. Lint, typecheck, e2e (37 passed) all clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **`exclude_extensions:` agent frontmatter — extension denylist for subagents** ([#94](https://github.com/tintinweb/pi-subagents/issues/94) — thanks [@ramhaidar](https://github.com/ramhaidar)). Applied after the `extensions:` include set; exclude wins, including over `tools: ext:` selectors (an excluded extension never loads, so its `ext:` reference becomes the usual orphan warning). The key use case: `extensions: true` + `exclude_extensions: pi-notify` — all extensions except a noisy one, without hand-maintaining an allowlist. Plain canonical names only (case-insensitive); paths, `*`, and unmatched names fire `extension-error:…` warnings (warn-not-abort, as with `extensions:` mismatches); `extensions: false` + an exclude warns that the exclude has no effect. **Not a sandbox:** excluded extensions' factory code still executes once during loading — exclusion suppresses handler binding and tool registration, not load-time side effects. The negation syntax `extensions: ["*", "!name"]` was deliberately rejected: an unquoted `!name` is a YAML tag and silently mis-parses.
+
 ## [0.10.1] - 2026-06-10
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -196,6 +196,7 @@ All fields are optional — sensible defaults for everything.
 | `display_name` | — | Display name for UI (e.g. widget, agent list) |
 | `tools` | all 7 | Which tools the agent can call. Built-in names (`read, grep, …`), `*` / `all` (all built-ins), `none`, and `ext:<extension>` / `ext:<extension>/<tool>` selectors for extension tools. See [Tool & extension scoping](#tool--extension-scoping) below |
 | `extensions` | `true` | Which extensions to load for the agent. `true` (all defaults), `false` (none), or an explicit list: `[mcp, "/abs/path.ts", "*"]`. See [Tool & extension scoping](#tool--extension-scoping) below |
+| `exclude_extensions` | — | Extension denylist applied after `extensions:` — exclude wins. Plain names only (case-insensitive), no paths or `*`. Useful with `extensions: true` to drop one extension (e.g. `pi-notify`) |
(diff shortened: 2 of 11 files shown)


ramhaidar:
Thanks a lot @tintinweb 🤩

PR #102, opened 2026-06-10
feat(RPC): SpawnOptions.cwd — run subagents in another directory, config stays with the parent project.

### LINKED ISSUE #96: [Feature request] Allow callers to pass through `cwd` in options
I use pi-subagents via RPC in another extension, and I need the ability to spawn a subagent in a certain directory, rather than the parent sessions cwd or temp worktree.

Had a quick poke around - `runAgent` already has a `cwd` override, but its not passed through from `AgentManager.startAgent` (which just sets `worktreeCwd`)

Proposed change, widening only:
1. Add `cwd?: string` to `SpawnOptions`
2. Use it in `startAgent`, `cwd: options.cwd ?? worktreeCwd`

It'd be a one liner so happy to open a PR if you're open to adding the option?

Thanks!

tintinweb:
## Summary

  Adds `SpawnOptions.cwd` — spawn a subagent in a different working directory than the parent session. Closes #96 (thanks @madeleineostoja for the report and diagnosis).

  For RPC/programmatic callers only — **not** exposed on the `Agent` tool, so the LLM-visible surface is unchanged. RPC callers get it for free since spawn options pass through verbatim.

  ## Design: the agent works *there*, but carries the parent's toolbox

  The naive version of this feature ("run the agent over there") would execute the target directory's `.pi` extensions in-process — equivalent to opening pi in an untrusted directory. Instead, this PR splits the two meanings of cwd:

  - **Working directory** (bash/read/write/edit, the prompt's environment block) → the caller's target.
  - **Config discovery** (`.pi` project extensions, skills, pi settings, agent memory) → the parent session's project, via a new `RunOptions.configCwd`.

  So the target's `.pi` extensions **never execute**, its agents/skills are never prompt content, and `memory:` agents keep reading/writing the parent project's memory. A caller who ever wants target-config semantics can get an explicit opt-in
  later; the safe behavior is the default.

## New API surface

  | Option | Where | Audience | Semantics |
  |---|---|---|---|
  | `cwd?: string` | `SpawnOptions` (agent-manager) | **Public** — RPC callers via `subagents:rpc:spawn` `options`, and direct `manager.spawn()` embedders | Absolute path to an existing directory. The agent's tools work there; `.pi` config still loads from the parent session's project. `null`/omitted = parent cwd (unchanged behavior). Invalid values (relative, missing, file, non-string) throw curated errors at spawn → RPC error envelopes. Composes with `isolation: "worktree"` (worktree created from the target repo, branch lands there). |
  | `configCwd?: string` | `RunOptions` (agent-runner) | **Internal** — set only by `AgentManager.startAgent`; not reachable through RPC or the Agent tool | Where `.pi` config is discovered (project extensions, skills, pi settings, agent memory).  Defaults to the working directory; the manager pins it to the parent session's cwd whenever a caller-supplied `cwd` is in play. JSDoc carries a warning for future callers: passing `cwd` without `configCwd` executes the target's `.pi` extensions — worktree isolation is the one intentional exception. |
  | `workPath: string` | `WorktreeInfo` / `AgentRecord.worktree` | **Internal** — produced by `createWorktree`, consumed by `startAgent` | The cwd-equivalent location *inside* the worktree copy (repo-root-relative, realpath-safe). Equals `path` when the worktree was created from the repo root. Used only for caller-supplied cwds; plain worktree spawns keep the historical root working dir. |

  Not added anywhere: no new `Agent` tool parameter (LLM schema unchanged), no new persisted setting, no RPC protocol change (`cwd` is an optional field in the existing `options` passthrough; absent = exact previous behavior).

## Usage example (RPC caller)

  ```typescript
  // From another pi extension: spawn an agent that works in /work/monorepo/packages/api,
  // isolated in a worktree, while carrying THIS session's .pi config (extensions/skills/agents).
  export default function (pi: ExtensionAPI) {
    pi.registerCommand("audit-api", async (_args, ctx) => {
      const requestId = crypto.randomUUID();

      const unsub = pi.events.on(`subagents:rpc:spawn:reply:${requestId}`, (reply) => {
        unsub();
        if (!reply.success) {
          // Validation errors land here, at the call:
          //   "SpawnOptions.cwd must be an absolute path: …"
          //   "SpawnOptions.cwd does not exist: …"
          //   "SpawnOptions.cwd is not a directory: …"
          ctx.ui.notify(`Spawn failed: ${reply.error}`, "error");
        } else {
          ctx.ui.notify(`Agent ${reply.data.id} started`, "info");
        }
      });

      pi.events.emit("subagents:rpc:spawn", {
        requestId,
        type: "general-purpose",            // resolved from the PARENT project's agent registry
        prompt: "Run the test suite and fix the failing tests.",
        options: {
          description: "fix api tests",
          run_in_background: true,
          cwd: "/work/monorepo/packages/api",  // ← new: absolute path, must exist
          isolation: "worktree",               // optional: worktree created from /work/monorepo,
                                               // agent works at packages/api INSIDE the copy
        },
      });

      // Completion arrives like any background agent: `subagents:completed` /
      // `subagents:failed` on the event bus. With worktree changes, the result ends:
      //   Changes saved to branch `pi-agent-<id>` in `/work/monorepo`.
      //   Merge with: `git merge pi-agent-<id>` (run in `/work/monorepo`)
    });
  }
  ```

  What the agent sees: its tools operate in `packages/api` (or the worktree copy of it) and the
  prompt's environment block describes that directory — but `.pi` extensions, skills, settings,
  agent types, and memory all come from the spawning session's project. The target directory's
  `.pi` is never loaded, so pointing agents at cloned/untrusted repos is safe by default.

  ## Composes with `isolation: "worktree"`

  - The worktree is created **from the target directory's repo**; the resulting `pi-agent-*` branch lands there, and the completion message names the repo (so the orchestrator doesn't `git merge` in the wrong one).
  - Subdirectory scoping survives isolation: a cwd of `/monorepo/packages/api` runs the agent at `packages/api` *inside the copy* (new `WorktreeInfo.workPath`, resolved via `git rev-parse --show-toplevel` with realpath on both sides for the macOS
  `/tmp` symlink case) instead of silently widening to the repo root.
  - On dispose, worktree registrations are pruned in **every** repo that received one (new tracking set) — only a hard crash can leave a stale entry.

  ## Validation: strict, typed, early — and re-checked

  `cwd` must be an absolute path to an existing directory. Non-strings (RPC callers send arbitrary JSON), relative paths, missing paths, and files each throw a curated error at `spawn()` — before queueing, so the RPC caller gets an error envelope
  at the call, not a confusing late failure. `null` is treated as unset (standard JSON for "no value"). Because queued spawns can start minutes later, the check re-runs at queue drain (TOCTOU).

  ## Normal users: provably unchanged

  - With `cwd` unset, every new code path collapses to the previous expressions; `configCwd` stays unset, so all config resolution is the pre-change value — including plain worktree isolation, which deliberately keeps its historical behavior (agent
  at the copy's root, config inside the copy). A dedicated test pins the subtlest case: plain worktree from a repo subdirectory must **not** adopt `workPath` (doing so would move `.pi` discovery and silently drop extensions/skills).
  - Empirical check: the **unmodified master test suite passes against this branch — 594/594, zero failures**.
  - Default-case result text, tool schema, and descriptions are byte-identical.

  ## Tests (11 new)

  - cwd → `runAgent` working dir + parent-anchored `configCwd`; no-cwd case pins both unset (asserted on `mock.lastCall`, not aggregate history)
  - `cwd: null` behaves exactly like omitted
  - worktree composition: created from target, session at `workPath`, cleanup against the target repo
  - plain-worktree historical behavior preserved even when `workPath` differs
  - validation: relative path, nonexistent path, regular file, non-string — each with its curated error and no orphan record
  - real-git `workPath` tests: equals root when created from the root; preserves `packages/api` scoping from a subdir

  Full suite: 605 passed / 4 skipped. Lint and typecheck clean. README (RPC spawn section incl. the trust model and worktree notes) and CHANGELOG updated.


madeleineostoja:
Thanks for the quick add! I'll test this today

madeleineostoja:
Works perfectly! Thanks @tintinweb 

Also tested with project/repo config, didn't see any issues

tintinweb:
I'll get this out before monday. thanks for the testing!

PR #105, opened 2026-06-14
feat: complete foreground agent lifecycle and add conversation logging

benrhodeland:
## Problem

Foreground agents (the default — `run_in_background` unset or `false`) produce **zero persistent logs**:
- No `.output` file with the conversation
- No `subagents:record` entry in the parent JSONL  
- No `subagents:completed`/`subagents:failed` lifecycle events
- The full subagent conversation is discarded after `spawnAndWait` returns

Background agents (`run_in_background: true`) get all of the above. This asymmetry means foreground subagent conversations are permanently lost, and cross-extension observers (OpenTelemetry, session auditors) see an orphaned lifecycle (`started` but never `completed`).

## Solution

Two independent fixes in one PR:

### 1. Fire `onComplete` for foreground agents (lifecycle symmetry)

In `agent-manager.ts`, both `.then()` and `.catch()` blocks now call `onComplete` for foreground agents with `resultConsumed = true`. This:
- Emits `subagents:completed`/`subagents:failed` events
- Writes `subagents:record` metadata entry to parent JSONL
- Skips all notification/batch/group-join logic (via the `resultConsumed` early return)

### 2. Stream foreground conversations to `.output` files

In `index.ts`, the foreground `onSessionCreated` callback now wires `streamToOutputFile` (same as background agents). The output file path is set via a new `onSpawned` hook in `spawnAndWait`, called synchronously after `spawn()` but before `onSessionCreated` fires.

### API change

`spawnAndWait` now returns `{ id, record }` instead of just `record`, so callers can access the agent ID for output file setup.

## Safety

- `resultConsumed = true` makes `onComplete` hit its early return — only lifecycle event + JSONL entry are emitted, no notifications
- Multiple sequential foreground agents: framework dispatches one-at-a-time, zero overlap
- All shared-state mutations are idempotent (`agentActivity.delete`, `widget.markFinished`)
- Foreground agents never touch `currentBatchAgents`, `groupJoin`, or `runningBackground`
- Test suite: 605 tests pass (1 test updated to expect new behavior)

## Files changed

- `src/agent-manager.ts` — onComplete for foreground, onSpawned hook, spawnAndWait return type
- `src/index.ts` — streamToOutputFile for foreground, output file setup via onSpawned
- `test/agent-manager.test.ts` — update foreground onComplete test

### DIFF
--- a/src/agent-manager.ts
+++ b/src/agent-manager.ts
@@ -311,7 +311,12 @@ export class AgentManager {
           }
         }
 
-        if (options.isBackground) {
+        // Fire onComplete for foreground agents too — lifecycle symmetry.
+        // Mark resultConsumed so the callback skips notifications (result returned inline).
+        if (!options.isBackground) {
+          record.resultConsumed = true;
+          try { this.onComplete?.(record); } catch { /* ignore completion side-effect errors */ }
+        } else {
           this.runningBackground--;
           try { this.onComplete?.(record); } catch { /* ignore completion side-effect errors */ }
           this.drainQueue();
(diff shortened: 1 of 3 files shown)


tintinweb:
thanks! 🙏

Issue #106, opened 2026-06-16
The model passed as param does not override the model from frontmatter


frabul:
If an agent front matter specifies a model then the model invocation parameter is ignored. That is not in line with README.md.
The problem is particularly awkward when the model from the agent definition does not exist, in that case the main agent model is taken as  fallback also if one was specified as parameter.

tintinweb:
Thanks for the report! The first behaviour is intentional and documented — from the README:

> **Frontmatter is authoritative.** If an agent file sets `model` [...] those values are locked for that agent. `Agent` tool parameters only fill fields the agent config leaves unspecified.

The second point is a fair edge case: when a frontmatter-pinned model fails to resolve, the fallback currently ignores any caller-supplied model and drops back to the parent. That's worth looking at separately.

PR #107, opened 2026-06-16
fix: inline provided 'model' param wins over the agent's frontmatter

frabul:
Address issue #106.


tintinweb:
WAD. template config (if set) is authorative.

PR #111, opened 2026-06-23
feat: persist_session + session_dir frontmatter for subagents

codesoda:
## Summary

Adds two optional frontmatter fields to custom agents so a subagent can run as a **real, persisted pi session** instead of an in-memory-only one:

- **`persist_session`** (default `false`) — when `true`, the subagent runs through `SessionManager.create(...)` so its transcript is written to pi's normal session location, exactly like a top-level session. When `false` (the default), behaviour is unchanged (`SessionManager.inMemory(...)`).
- **`session_dir`** (optional) — overrides where the persisted session is written when `persist_session: true`. Supports absolute paths, `~`, and cwd-relative paths. Omitted ⇒ pi's normal session location (via `PI_CODING_AGENT_SESSION_DIR` or the settings manager's `getSessionDir()`).

## Motivation

When orchestrating long-running, multi-round workflows (plan → review → implement → verify), it's valuable to have each subagent's full conversation persisted as a normal pi session — so it shows up in `~/.pi/agent/sessions`, can be inspected after the fact, and can be resumed. Today subagents are in-memory only. This makes persistence opt-in per agent via frontmatter, with zero change to the default.

## Changes

- `src/types.ts` — add `persistSession?: boolean` and `sessionDir?: string` to `AgentConfig`.
- `src/custom-agents.ts` — parse `persist_session` / `session_dir` from frontmatter.
- `src/agent-runner.ts` — resolve the configured session dir (absolute / `~` / cwd-relative) and select `SessionManager.create(...)` vs `SessionManager.inMemory(...)` based on `persistSession`; default session dir falls back to `PI_CODING_AGENT_SESSION_DIR` then `settingsManager.getSessionDir?.()`.
- `README.md` — document both fields in the frontmatter table.
- Tests — `test/agent-runner.test.ts` (in-memory default, persisted to pi's normal location, frontmatter `session_dir` override) and `test/custom-agents.test.ts` (frontmatter parsing for both fields).

## Compatibility

Fully backward compatible — defaults preserve the existing in-memory behaviour. The sidechain output transcript is still written either way.

## Verification

- `npm run typecheck` — clean
- `npm run lint` — clean (exit 0)
- `npm run test` — added unit tests pass; full unit suite green

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,8 @@ All fields are optional — sensible defaults for everything.
 | `model` | inherit parent | Model — `provider/modelId` or fuzzy name (`"haiku"`, `"sonnet"`) |
 | `thinking` | inherit | off, minimal, low, medium, high, xhigh |
 | `max_turns` | unlimited | Max agentic turns before graceful shutdown. `0` or omit for unlimited |
+| `persist_session` | `false` | Persist this subagent as a normal pi session instead of keeping the session in memory only. The sidechain output transcript is still written either way |
+| `session_dir` | pi default | Optional session directory when `persist_session: true`; omitted uses pi's normal session location, and relative paths resolve from the agent cwd |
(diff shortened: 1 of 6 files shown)


tintinweb:
great work, thanks! 🙌

codesoda:
@tintinweb  No worries, anytime

PR #113, opened 2026-06-23
feat: emit session_ready & tools_resolve lifecycle events

NovusEdge:
## Summary

Add two new lifecycle events that enable extensions to integrate properly with subagent sessions:

1. **`subagents:session_ready`** - Fires when an agent's session is created and ready
2. **`subagents:tools_resolve`** - Fires before session creation to allow tool injection

## Motivation

Extensions like memory/context systems need to:
1. Attach state to the subagent's session when it starts
2. **Inject additional tools** (e.g., memory tools) into subagent sessions
3. Access the session object to inject context or track activity
4. Clean up when the agent completes

Currently, the existing events (`subagents:started`, `subagents:completed`) don't expose the session object or allow tool injection, making this impossible without relying on internal APIs.

## Changes

- Emit `subagents:session_ready` event in `agent-manager.ts` right after `onSessionCreated` fires
- Emit `subagents:tools_resolve` event in `agent-runner.ts` before session creation
- Add `onToolsResolve` callback option to `SpawnOptions` and `RunOptions`
- Update README to document the new events

### session_ready
```typescript
{
  id: string;            // agent ID
  type: string;          // subagent type (Explore, Plan, etc.)
  session: AgentSession; // the session object
  record: AgentRecord;   // full record for metadata
}
```

### tools_resolve
```typescript
{
  type: string;    // subagent type
  tools: any[];    // mutable array - push to inject tools
  agentId: string; // agent ID
}
```

### Context propagation (session_ready)
```typescript
pi.events.on("subagents:session_ready", ({ id, type, session, record }) => {
  const childHarness = parentHarness.fork({ mode: 'fork', tagPrefix: type });
  session.veilHarness = childHarness;
  trackForMerge(id, childHarness);
});
```

### Tool injection (tools_resolve)
```typescript
pi.events.on("subagents:tools_resolve", (event) => {
  const veilTools = parentHarness.getTools();
  event.tools.push(...veilTools);
});
```

## Backward Compatibility

This is purely additive - existing code is unaffected. Extensions can opt-in to the new events.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ https://github.com/user-attachments/assets/8685261b-9338-4fea-8dfe-1c590d5df543
 - **Skill preloading** — inject named skills into agent system prompts, discovered from `.pi/skills/`, `.agents/skills/`, and global locations (Pi-standard `<name>/SKILL.md` directory layout supported)
 - **Tool denylist** — block specific tools via `disallowed_tools` frontmatter
 - **Styled completion notifications** — background agent results render as themed, compact notification boxes (icon, stats, result preview) instead of raw XML. Expandable to show full output. Group completions render each agent individually
-- **Event bus** — lifecycle events (`subagents:created`, `started`, `completed`, `failed`, `steered`, `compacted`) emitted via `pi.events`, enabling other extensions to react to sub-agent activity
+- **Event bus** — lifecycle events (`subagents:created`, `started`, `session_ready`, `completed`, `failed`, `steered`, `compacted`) emitted via `pi.events`, enabling other extensions to react to sub-agent activity. The `session_ready` event exposes the full `AgentSession` object for extensions that need to attach state (e.g., memory systems, context managers)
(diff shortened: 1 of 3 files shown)


tintinweb:
I think I need to look a little deeper into this. for example the tools injection. question is if this should be done on subagent creating. also, do we want to all all connected extensions to mess with the subagents. `session_ready` might be easier to land. also we should probably just emit copies of the session/record. I want to be a bit careful on how low-level connected extensions can go. 🤔

NovusEdge:
Hmm fair point. I think I have a proposal in that case.

For `session_ready`, I can update it to emit frozen copies instead of live objects. That should be safe to land.

For `tools_resolve`, I do need something like it. [veil](https://github.com/engrammic-ai/veil) needs to inject context tools into subagents so they share memory with the parent. 

Buuut I do agree, the mutable array is too open. Some options that come to mind to fix this:

1. Spawn-time opt-in + add-only: only fire if spawner passes `allowToolInjection: true`, and extensions can only add tools, not remove or shadow
2. Split the PR, land `session_ready` first while we figure out the right shape for tool injection
3. Something better you have in mind?

PR #117, opened 2026-06-26
fix(ui): avoid duplicate foreground agent rendering

### LINKED ISSUE #118: Foreground agent status can render twice and leave stale rows
## Bug description

When running a foreground `Agent` invocation in Pi, the same agent status can appear twice: once in the inline tool result and again in the persistent `Agents` widget. In terminal multiplexers such as tmux/zellij, rapid partial updates can also leave stale status rows behind until the UI is forced to redraw.

## Steps to reproduce

1. Run Pi inside tmux or zellij.
2. Start a foreground subagent, for example an `sdd-proposal`/custom agent without `run_in_background`.
3. Watch the running Agent status update while tool-use counts change.
4. Optionally press Ctrl+O or zoom/unzoom the terminal pane to force a redraw.

## Expected behavior

- Foreground agents should render only in the inline tool result.
- The persistent `Agents` widget should show background agents only.
- Partial updates should replace previous status rows cleanly.

## Actual behavior

- The foreground agent can appear in both places.
- Old rows such as previous tool-use counts can remain visible until a forced redraw.

## Proposed fix

PR #117 updates the widget to filter for background agents and changes the running Agent status renderer to use separate component lines instead of one multiline `Text` component.

## Environment

Observed with Pi running in tmux/zellij on macOS.

## Related PR

#117

Alan-TheGentleman:
Fixes #118

## Summary
- Keep the persistent `Agents` widget focused on background agents so foreground runs are not shown twice while their inline `Agent` tool result is already visible.
- Render the running foreground status as separate component rows instead of one multiline `Text` value.
- Add regression coverage for foreground filtering, background widget rendering, and the running-status component shape.

## Why
Foreground agents have two possible UI surfaces: the inline tool result and the persistent widget. Showing the same foreground run in both places makes the UI look duplicated. During fast live updates, keeping the running status as distinct component rows also avoids relying on multiline text replacement for the status block.

This was easiest to notice inside tmux/zellij, but the main fix is scoped to `pi-subagents` ownership: foreground runs should stay inline; background runs should stay in the widget.

## Test plan
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun run test test/agent-widget.test.ts`
- [ ] `bun run test` — currently fails in `test/subagents-print-mode-e2e.test.ts` because two print-mode E2E cases return `No output` instead of the expected child markers; unrelated to this UI widget/render change.

### DIFF
--- a/src/agent-manager.ts
+++ b/src/agent-manager.ts
@@ -167,6 +167,12 @@ export class AgentManager {
       abortController,
       lifetimeUsage: { input: 0, output: 0, cacheWrite: 0 },
       compactionCount: 0,
+      // Raw tri-state (not coerced to a boolean): true = background, false =
+      // foreground (has an inline tool-result surface), undefined = caller never
+      // declared it (e.g. a cross-extension RPC spawn). The widget's background-
+      // only filter excludes only explicit `false`, so undefined agents — which
+      // have no inline surface — stay visible instead of vanishing.
+      isBackground: options.isBackground,
       invocation: options.invocation,
     };
     this.agents.set(id, record);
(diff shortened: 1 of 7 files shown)


tintinweb:
thanks!

 +configurable widget (show all, background only, no widget)

PR #127, opened 2026-07-07
fix: get_subagent_result wait:true doesn't wait for queued agents

benrhodeland:
## Summary

`get_subagent_result(wait: true)` only waits when `record.status === "running" && record.promise`. A **queued** background agent (concurrency limit reached, `maxConcurrent`) has neither yet — `status` is `"queued"` and `promise` isn't assigned until the queue actually starts the agent (see `AgentManager.startAgent`). So calling `wait: true` on a queued agent skips the wait branch entirely and immediately returns "Agent is still running." — forcing the caller into a manual poll loop against its own queue.

This is the opposite of the documented pattern: the tool's guidance elsewhere says not to poll running agents, but there's no way to actually block on a queued one.

## Fix

Extend the wait condition to include `"queued"`, and poll (250ms) until the record leaves the queued state before falling through to the existing `await record.promise`. `resultConsumed` is still pre-marked before the wait so the completion notification path is unaffected.

```ts
if (params.wait && (record.status === "running" || record.status === "queued")) {
  record.resultConsumed = true;
  cancelNudge(params.agent_id);
  while (record.status === "queued") {
    await new Promise((r) => setTimeout(r, 250));
  }
  if (record.promise) await record.promise;
}
```

## Test plan

- [x] `test/wait-queued.test.ts` — wiring test through the real extension: spawns background agents past the concurrency limit, calls `get_subagent_result(wait: true)` on the queued one, drains the running agents, asserts the call returns the final result instead of "still running"
- [x] `npm run lint` / `npm run typecheck` / `npm run test` / `npm run build` all pass

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1362,10 +1362,15 @@ Terse command-style prompts produce shallow, generic work.
       // Pre-mark resultConsumed BEFORE awaiting: onComplete fires inside .then()
       // (attached earlier at spawn time) and always runs before this await resumes.
       // Setting the flag here prevents a redundant follow-up notification.
-      if (params.wait && record.status === "running" && record.promise) {
+      // Queued agents have no promise yet (it's created when the queue starts
+      // them), so poll until they leave the queue, then await like a running one.
+      if (params.wait && (record.status === "running" || record.status === "queued")) {
         record.resultConsumed = true;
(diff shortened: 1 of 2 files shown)


tintinweb:
thanks! 🤗

Issue #126, opened 2026-07-07
Child agent sessions never fire session_shutdown — per-spawn AgentManager/interval leak


benrhodeland:
## What happens

Every subagent spawn re-activates this extension inside the child's own agent session (`runAgent` → `createAgentSession` → `session.bindExtensions()` in `src/agent-runner.ts`). Each re-activation runs the whole default-export function again, which constructs a brand-new `AgentManager` (`src/index.ts`, `new AgentManager(...)` near the top of the export). That constructor starts an unref'd `setInterval(() => this.cleanup(), 60_000)`.

That interval — and the whole child `AgentManager` instance, including its `worktreeRepos` set and its own copy of the `Symbol.for("pi-subagents:manager")` registry claim — is only ever torn down by that same activation's own `session_shutdown` handler, which calls `manager.dispose()` (clears the interval, disposes worktrees, clears the agents map).

The problem: when a subagent's session is later cleaned up (via the *parent's* manager, e.g. `removeRecord`/`cleanup()` calling `record.session?.dispose()`), that only calls the child `AgentSession`'s own `dispose()` — which invalidates the extension context and disconnects listeners, but does **not** emit `session_shutdown` to the child's bound extensions. I checked this against `node_modules/@earendil-works/pi-coding-agent/dist/core/agent-session.js`: `session_shutdown` is only emitted on `reload` (line ~1908); the plain `dispose()` method (line ~469) does not emit it.

So the child activation's `session_shutdown` handler — the only thing that clears its interval and releases its registry claim — never runs for an ordinary subagent completion (success, error, or abort). Every subagent you spawn leaves behind one orphaned `AgentManager` object and one dangling 60-second interval for the life of the host process.

## Why it's low-severity but worth tracking

- The interval is `.unref()`'d, so it doesn't keep the process alive.
- It's bounded per spawn (not unbounded growth from a single long-running agent).
- No data/state corruption — the child manager never holds records anyone else reads (see #128 , which fixes a related but distinct registry-clobber bug).

It's still a real, indefinite accumulation of live timers + retained objects in any process that spawns many subagents over its lifetime (long print-mode/headless runs, CI agents, cron-driven sessions) — bounded by "how many subagents this process ever spawns," not by anything that gets cleaned up.

## Where I think the fix has to live

[...]

PR #128, opened 2026-07-07
fix: don't let re-activations clobber the Symbol.for manager registry

benrhodeland:
## Summary

Subagent sessions re-activate this extension in the same process (`session.bindExtensions` inside `runAgent`, `src/agent-runner.ts`) — every subagent spawn runs the whole extension's activation code again inside the child's own session. The `Symbol.for("pi-subagents:manager")` global registry didn't account for that: every activation unconditionally overwrote the slot with its own manager, and every activation's `session_shutdown` unconditionally deleted it.

Net effect: after the first subagent in a session completes, `globalThis[Symbol.for("pi-subagents:manager")]` points at that child's (usually already-shutting-down) manager instead of the root session's — and the child's own shutdown then deletes the slot entirely, even though the root session is still alive and may still have running background agents.

This isn't hypothetical — `test/helpers/print-mode-runner.ts` in this repo documents a real cross-package consumer: headless/print-mode hosts (the `pi-chonky-step` pattern) poll this exact global's `hasRunning()`/`waitForAll()` to decide whether to keep the process alive while background subagents finish. Under the old code, that check can report `false`/empty after any single subagent completes, even while the root session has real background agents still running — a headless host could exit and kill them mid-run.

## Fix

First activation to see an empty slot claims it (`ownsManagerRegistry`); later activations (children) leave it alone. Only the owning activation's shutdown releases it, and only if the slot still holds *that* activation's own entry (identity check), so a stale/superseded owner's shutdown can't delete a successor's live entry.

```ts
const ownsManagerRegistry = (globalThis as any)[MANAGER_KEY] === undefined;
if (ownsManagerRegistry) {
  (globalThis as any)[MANAGER_KEY] = registryEntry;
}
// ...on session_shutdown:
if (ownsManagerRegistry && (globalThis as any)[MANAGER_KEY] === registryEntry) {
  delete (globalThis as any)[MANAGER_KEY];
}
```

## Test plan

- [x] `test/manager-registry-guard.test.ts` — drives two real activations of the extension through the real `session_shutdown` lifecycle: spawns a background agent through the root, activates a second ("child") instance, asserts the registry still resolves to the root's manager and the root's record, asserts the child's shutdown does not delete the slot, and asserts the root's own shutdown does release it
- [x] `npm run lint` / `npm run typecheck` / `npm run test` / `npm run build` all pass

## Related

While tracing this I found a second, related but distinct issue in the same area — child agent sessions never fire `session_shutdown` at all on ordinary completion (only on `reload`), so each child's own `AgentManager` (interval, worktree tracking) is never disposed. That one can't be fixed from this side of the boundary (needs a `pi-coding-agent` teardown hook), so I filed it separately as #126 rather than folding it into this PR.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -438,14 +438,25 @@ export default function (pi: ExtensionAPI) {
 
   // Expose manager via Symbol.for() global registry for cross-package access.
   // Standard Node.js pattern for cross-package singletons (used by OpenTelemetry, etc.).
+  //
+  // Claim the slot only if it's free: subagent sessions re-activate this
+  // extension in the same process (session.bindExtensions in agent-runner.ts),
+  // and unconditionally overwriting would point the registry at a short-lived
+  // child manager — and the child's shutdown would then delete the root
+  // session's entry. The first activation (the root session) wins; child
+  // activations leave it alone.
   const MANAGER_KEY = Symbol.for("pi-subagents:manager");
-  (globalThis as any)[MANAGER_KEY] = {
+  const registryEntry = {
     waitForAll: () => manager.waitForAll(),
     hasRunning: () => manager.hasRunning(),
     spawn: (piRef: any, ctx: any, type: string, prompt: string, options: any) =>
       manager.spawn(piRef, ctx, type, prompt, options),
     getRecord: (id: string) => manager.getRecord(id),
   };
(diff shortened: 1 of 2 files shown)


tintinweb:
thanks! 🤗

PR #129, opened 2026-07-08
update pi dev dependency to 0.80.x

philipmw:
The committed package-lock pinned @earendil-works/pi-* to 0.75.5 (the latest published when the lock was generated), so `npm install` resolved the peer dependencies to 0.75.5 even though the peer range (>=0.74.0) and the registry now allow 0.80.3 — the version pi actually ships and runs. Dev/test/typecheck were therefore exercising an older API surface than production, which silently hid behavior differences (e.g. createAgentSession's `excludeTools` option, which does not exist in 0.75.5).

Regenerate package-lock.json so the peers resolve to 0.80.3, matching the runtime. Accommodate the one breaking change this surfaces in the test harness: pi-ai moved its faux/model test helpers (registerFauxProvider, getModel, fauxToolCall, Model, ToolCall, Context) from the package root to the `/compat` subpath in 0.80.x, so repoint those imports.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,21 +46,6 @@
         }
       }
     },
-    "node_modules/@aws-crypto/crc32": {
-      "version": "5.2.0",
-      "resolved": "https://registry.npmjs.org/@aws-crypto/crc32/-/crc32-5.2.0.tgz",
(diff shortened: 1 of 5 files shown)


philipmw:
This PR is not just for funsies but because I have another PR in the wings that fixes a bug but relies on the newer dependency. @tintinweb please ship!

tintinweb:
thanks! I updated this with #140

PR #134, opened 2026-07-09
Support explicit subagent session files

### LINKED ISSUE #131: Support explicit session file paths for persistent subagents
## Summary

Please support an explicit subagent session file path/name, not only `persist_session: true` plus `session_dir`.

Today `persist_session` can write a subagent as a normal Pi session and `session_dir` can choose the directory, but the actual JSONL filename is still generated by Pi's `SessionManager`. For larger multi-task workflows it would be useful to make the file path deterministic and user-managed.

## Motivation

When coordinating many tasks, I often want stable semantic subagent sessions such as:

```text
.agents/sessions/KEY.dev.jsonl
.agents/sessions/KEY.reviewer.jsonl
.agents/sessions/KEY.plan.jsonl
```

This makes the session file itself part of the task workflow:

- the parent/orchestrator can intentionally assign a durable lane to a subagent;
- humans can later inspect, resume, or reply to the same session by path;
- external task dashboards can map issue keys and roles to known files without guessing generated session IDs;
- background agent management becomes much easier when many Linear/GitHub/GitLab tasks are active at the same time;
- repeated plan/review/dev loops can reuse a stable context instead of creating many unrelated generated session files.

## Desired API

One possible design is to add a frontmatter field, for example:

```yaml
---
persist_session: true
session_file: .agents/sessions/KEY.reviewer.jsonl
---
```

And/or support an `Agent` tool parameter for dynamic task-specific use:

```js
Agent({
  subagent_type: "reviewer",
  prompt: "Review the current implementation",
  description: "Review KEY",
  run_in_background: true,
  session_file: ".agents/sessions/KEY.reviewer.jsonl",
})
```

The exact name is not important; `session_file`, `session_path`, or another explicit term would be fine.

## Expected behavior

- Relative paths should resolve from the agent cwd, similar to `session_dir`.
- `~` and absolute paths should be supported if they are acceptable for `session_dir`.
- The feature should probably imply `persist_session: true`, or produce a clear validation error if persistence is not enabled.
- If the target file already exists, the behavior should be explicit: either resume/append to it, or fail unless a separate resume/overwrite flag is supplied. Silent overwrite should be avoided.
- The resulting session should still work as a normal Pi session JSONL and remain compatible with `session_dir` defaults when `session_file` is omitted.

## Current workaround

Use `persist_session: true` and point `session_dir` at a task-specific directory, then discover the generated filename afterward. This helps but does not solve the stable lane problem, because the filename is not predictable.

## Related existing feature

This builds on PR #111 (`persist_session` / `session_dir`). The request here is more specific: choose the concrete JSONL session file so humans and task automation can continue the same subagent lane later.

wenerme:
## Summary

- Add `session_file` to custom agent frontmatter, the `Agent` tool parameters, scheduled subagents, and spawn/run configuration.
- Resolve relative `session_file` paths from the requested agent cwd before worktree isolation, so stable project session lanes do not get written into temporary worktrees.
- Open explicit session files through `SessionManager.open(...)` for append/resume, create missing parent directories, and treat `session_file` as implying persistence.
- Document the feature and add focused filesystem coverage.

Closes #131.

## Validation

- `./node_modules/.bin/biome check src/ test/`
- `npm run typecheck`
- `npx vitest run test/agent-runner.test.ts test/agent-manager.test.ts test/custom-agents.test.ts test/invocation-config.test.ts test/session-file.test.ts`
- `npm run test`
- `git diff --check`

Note: `npm run lint` returned `ESLint output (JSON parse failed: EOF...)` in this local harness, with no Biome diagnostics. The script's underlying command, `biome check src/ test/`, passed directly.


wenerme:
@tintinweb This PR is now rebased onto the latest `master` (`e955e29`), with a single feature commit (`9e99d41`). All three CI checks pass.

The remaining use case from #131 is choosing a stable JSONL path such as `.agents/sessions/KEY.reviewer.jsonl` for task/role-based continuation across parent sessions and external tooling. `rememberAgents` and `@handle` cover automatic persistence and resume, but do not let callers choose that path. Concurrent runs against the same file are explicitly unsupported.

Could you review whether this `session_file` API fits upstream, or whether you would prefer a narrower interface?

PR #135, opened 2026-07-10
fix: approve transitive build scripts so pnpm install exits 0

cad0p:
## Problem

With pnpm 11 as the configured package manager (`"npmCommand": ["pnpm"]` in pi settings), installing this package as a pi git package fails to load:

```
Error: Failed to load extension ".../tintinweb-pi-subagents/src/index.ts": Failed to load extension: Cannot find module 'croner'
Require stack:
- .../tintinweb-pi-subagents/src/schedule.ts
```

## Root cause

pi reconciles git packages with `git clean -fdx` then `<npmCommand> install`. Two transitive dependencies pulled in through the `@earendil-works/pi-*` peers ship native build scripts: `@google/genai` and `protobufjs`.

pnpm 11 blocks unapproved build scripts by default and exits non-zero (`[ERR_PNPM_IGNORED_BUILDS]`) when any are present. So `pnpm install` returns a failing exit code even though the dependencies themselves are installed correctly. pi's `runCommand` treats the non-zero exit as a failure, which breaks reconciliation and leaves the extension unable to load.

```
$ pnpm install
+ croner 10.0.1
[ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: @google/genai@1.52.0, protobufjs@7.6.5
$ echo $?
1
```

## Fix

Add a `pnpm-workspace.yaml` with an `allowBuilds` entry — the pnpm 11 replacement for the removed `onlyBuiltDependencies` setting — approving the two transitive packages with build scripts. `pnpm install` now exits 0.

```yaml
allowBuilds:
  "@google/genai": true
  protobufjs: true
```

The file is tracked, so it survives pi's `git clean -fdx` reconcile step.

## Verification

```
$ git clean -fdx        # simulate pi reconcile
$ pnpm install
+ croner 10.0.1
$ echo $?
0
$ node -e "require('croner'); console.log('croner OK')"
croner OK
```

- Before: `pnpm install` exits 1 (`ERR_PNPM_IGNORED_BUILDS`).
- After: exits 0, `croner` resolves, extension loads.
- `npm run lint` and `npm run typecheck` pass.
- The 10 pre-existing test failures are unchanged by this file (verified by stashing) — config-only change.


cad0p:
Closing — opening against my fork instead.

Issue #125, opened 2026-07-06
unable to use lazily loaded extension tools in subagents


CovetingEpiphany2152:
I have an issue where my subagents are unable to use tools from [context-mode](https://github.com/mksglu/context-mode/).

Despite adding the following fields in my frontmatter,

```yaml
  extensions: true
  tools: "*, ext:extension/ctx_execute"
  skills: context-mode
```

The agent is unable to find the `ctx_execute` tool.

From my understanding,

- pi-subagents computes extension tools before creating the session
- However, context-mode does not register tools during extension load. It [starts the MCP bridge on before_agent_start](https://github.com/mksglu/context-mode/blob/f267bdae970f0b01652a2ab413793318ed02065a/src/adapters/pi/extension.ts#L635)
  - Which then does [pi.registerTool()](https://github.com/mksglu/context-mode/blob/f267bdae970f0b01652a2ab413793318ed02065a/src/adapters/pi/mcp-bridge.ts#L1023)

From what I could find, Pi loads extensions in paths that may never run an agent turn: help/version, package commands, config, trust probing, etc. Eagerly spawning the context-mode MCP bridge during extension discovery caused orphan child processes / hangs in those non-agent paths. `before_agent_start` is the first reliable signal that Pi is actually about to make a model call, so context-mode starts the MCP bridge there. Pi supports this: pi.registerTool() can be called after startup, and tools refresh immediately in-session.

Is there any way to allow tools registered after loading extension to be used in subagents?

PR #141, opened 2026-07-11
fix: let subagents see asynchronously-registered extension tools (MCP)

philipmw:
Subagents could not use tools provided by MCP extensions (e.g. pi-mcp-extension), even though those tools work in the main agent. pi-mcp registers its tools asynchronously from its session_start handler, after connecting to MCP servers — never during synchronous extension load. runAgent, however, snapshotted extension.tools.keys() into a static `tools:` allowlist right after loader.reload() and before bindExtensions() fires session_start. In pi-mono that allowlist (allowedToolNames) is frozen at construction and gates the tool *registry* itself, so MCP tools — registered later — were dropped permanently, not just deactivated.

Scope extension tools via the session's live denylist instead of a frozen allowlist, matching how the main agent stays in sync. When extensions load and there is no `ext:` narrowing, pass `tools: undefined` (allowlist unset, so the live isAllowedTool gate admits tools whenever they register) and express scope as `excludeTools` = this extension's own orchestration tools + built-ins the agent didn't ask for + disallowedTools. Because an unset allowlist only activates pi's four default built-ins at turn 1, repair the active set post-bind via setActiveToolsByName(getAllTools()); tools registered later (lazy MCP servers) auto-activate through pi's refresh path. The static-allowlist path is retained unchanged for noExtensions/isolated and for `ext:` narrowing, where the exact set is known up front.

May solve issue #125.

I (the human) tested this manually by setting my local checkout of pi-subagents as the extension, then asking the LLM to spawn a subagent and use a search tool. Previously, it couldn't -- and now it can.

### DIFF
--- a/src/agent-runner.ts
+++ b/src/agent-runner.ts
@@ -547,20 +547,57 @@ export async function runAgent(
     }
   }
 
-  // Build the master tool allowlist applied at session construction.
-  // pi-mono's `allowedToolNames` gates BOTH registration and the initial active
-  // set, so listing the exact final set here means the session is correctly
-  // scoped from the first instant — no post-construction narrowing required.
+  // Tool scoping runs in one of two modes.
+  //
+  //  • Dynamic denylist (default whenever extension tools flow freely — i.e.
+  //    extensions load and there is no `ext:` narrowing). Some extensions register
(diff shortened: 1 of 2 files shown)


tintinweb:
Superseded by #157, which merged this fix (thanks @philipmw — credited in the changelog). Closing as resolved.

PR #146, opened 2026-07-12
feat: allow agents to suppress sidechain transcripts

Thoughts-One:
## Summary
- add `output_transcript: false` custom-agent frontmatter
- skip sidechain path creation, initial writes, and streaming for transcript-disabled foreground and background agents
- preserve transcript creation by default and document independence from `persist_session`

## Why
Some agents may process data that is permitted in memory for immediate operation but must not be copied into a durable sidechain transcript. `persist_session: false` already keeps the Pi session in memory, but the separate `.output` transcript remains unconditional.

## Verification
- `npm test` — 685 passed, 4 skipped
- `npm run typecheck`
- `npm run lint`
- wiring tests assert path creation, initial writes, and stream binding are all suppressed for foreground and background agents, while default behavior remains enabled

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ Individual agent results render Claude Code-style in the conversation:
 
 Completed results can be expanded (ctrl+o in pi) to show the full agent output inline.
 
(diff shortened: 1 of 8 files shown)


tintinweb:
. + a project wide setting switch + /agents -> Settings menu entry. frontmatter wins if set.

PR #168, opened 2026-07-23
fix(ui): show effective subagent runtime model

xz-dev:
## Summary

- retain the effective canonical `provider/modelId` for Agent UI rendering, including inherited and scheduled runs
- pair it with the child session's actual thinking level after Pi applies defaults and model capability clamping
- show that runtime metadata in running, queued, and finished Agent widgets, collapsed and expanded Agent results, and ConversationViewer
- keep resumed Agent results bound to the existing child session's current runtime metadata
- keep model identifiers safe for single-line terminal rendering

## Why

A thinking level such as `max` has little meaning without the model it applies to. The UI previously hid the model whenever a subagent inherited the parent model, and several background paths still omitted it entirely. Some paths also displayed requested configuration rather than the child session's effective model and thinking level.

The runtime Agent session is now authoritative once it exists. Before session creation, the immediate background result truthfully shows the resolved pre-session model metadata; later views read the live session state.

## Tests

- `npm run lint`
- `npm run typecheck`
- `npm run test` — 828 passed, 5 skipped
- `npm run build`

Coverage includes foreground/background Agent results, running/queued/finished AgentWidget rows, scheduled Agent metadata, ConversationViewer live runtime state, and resume after the existing session model/thinking changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ Restrictions:
 
 ## UI
 
-The extension renders a persistent widget above the editor showing active agents. By default it shows background runs only (`widgetMode: background`) — foreground agents already render inline as the `Agent` tool result, so the widget would otherwise double-render them. Switch to `all` (every agent) or `off` (hide the widget) via `/agents → Settings → Widget`:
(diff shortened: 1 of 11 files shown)


stanleytejakusuma:
hit this exact issue independently. spawned several background subagents configured on the same model as the parent session, but the running agent silently omitted the model name directly. so there's no way to tell what was actually driving a given subagent from the widget alone.

would genuinely like to see this merged - thanks for the fix!

tintinweb:
closed with #257. opt-in via setting `showModel`

PR #170, opened 2026-07-23
fix: shut down child sessions before disposal

### LINKED ISSUE #126: Child agent sessions never fire session_shutdown — per-spawn AgentManager/interval leak

liu-qingyuan:
## Summary

- emit and await `session_shutdown` before disposing retained child `AgentSession` instances
- apply the teardown contract to timed eviction, `clearCompleted()`, and manager shutdown
- detach records before awaiting teardown so concurrent cleanup paths cannot shut down the same session twice
- finish every child teardown with `Promise.allSettled`; log runner-level failures without blocking sibling or parent lifecycle cleanup

## Problem

`runAgent()` creates a standalone session and binds extensions, but `AgentSession.dispose()` only invalidates the extension context. It does not emit `session_shutdown`. The parent manager therefore discarded retained child sessions without giving child extensions a chance to release timers, transports, or subprocesses.

The unref'd manager interval described in #126 is one symptom. A child extension with a referenced MCP transport can also keep a settled headless process alive indefinitely.

## Design

`AgentManager` already owns child retention for resume and record eviction, so teardown stays private to that module. This mirrors the ordering used by Pi's `AgentSessionRuntime.dispose()` without migrating subagents to the larger runtime/session-replacement interface.

Completed sessions remain available for resume until the existing cleanup conditions remove their records. Production lifecycle handlers now await cleanup before proceeding.

## Verification

- `npm run lint`
- `npm run typecheck`
- `npm run test` — 752 passed, 5 skipped
- `npm run test:e2e` — 47 passed, 5 skipped
- `npm run build`
- real Pi 0.81.1 probe with an auto-connected Xcode MCP child: parent reached `agent_settled`, exited naturally with code 0, and left no process after 3 seconds

Closes #126

### DIFF
--- a/src/agent-manager.ts
+++ b/src/agent-manager.ts
@@ -24,6 +24,28 @@ export type CompactionInfo = { reason: "manual" | "threshold" | "overflow"; toke
 /** Default max concurrent background agents. */
 const DEFAULT_MAX_CONCURRENT = 4;
 
+/** Shut down and dispose every child session, logging failures after all settle. */
+async function shutdownAgentSessions(sessions: AgentSession[]): Promise<void> {
+  const outcomes = await Promise.allSettled(
+    sessions.map(async (session) => {
+      try {
+        await session.extensionRunner.emit({ type: "session_shutdown", reason: "quit" });
+      } finally {
+        session.dispose();
+      }
+    }),
+  );
+  const failures = outcomes
+    .filter((outcome): outcome is PromiseRejectedResult => outcome.status === "rejected")
+    .map((outcome) => outcome.reason);
+  if (failures.length > 0) {
+    console.warn(
+      `[pi-subagents] Failed to shut down ${failures.length} child agent session(s):`,
+      failures,
+    );
+  }
+}
+
 /**
  * Validate a caller-supplied SpawnOptions.cwd. `undefined`/`null` mean "unset"
  * (parent cwd). Anything else must be an absolute path to an existing
@@ -124,7 +146,11 @@ export class AgentManager {
     this.onCompact = onCompact;
     this.maxConcurrent = maxConcurrent;
     // Cleanup completed agents after 10 minutes (but keep sessions for resume)
-    this.cleanupInterval = setInterval(() => this.cleanup(), 60_000);
+    this.cleanupInterval = setInterval(() => {
+      void this.cleanup().catch((error) => {
+        console.warn("[pi-subagents] Failed to clean up completed agent sessions:", error);
+      });
+    }, 60_000);
     this.cleanupInterval.unref();
   }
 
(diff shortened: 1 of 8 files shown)


tintinweb:
fixed with https://github.com/tintinweb/pi-subagents/commit/9afe114c446c35804f3ca218bbc754cfb9620a95 ; credits in changelog, thx!.

PR #171, opened 2026-07-23
feat: support pi-subagents-style JSON agent overrides

ESSO0428:
## Summary

This PR adds support for configuring existing agents and registering new agents through Pi's `settings.json` files using the same `subagents.agentOverrides` format as [`npm:pi-subagents`](https://github.com/nicobailon/pi-subagents).

JSON overrides make agent configuration easier to adjust temporarily on a global or per-project basis. Users can quickly switch an agent's model or thinking level according to the current task, available provider quota, or cost constraints without editing or duplicating its complete `.md` definition.

Fields such as `model`, `thinking`, `systemPrompt`, `tools`, and `skills` can be configured under:

```text
subagents.agentOverrides.<agentName>
```

Local project settings can override only the fields that need to change while inheriting the remaining configuration from global settings or existing agent definitions.

When an overridden agent name does not already exist as a built-in or `.md` definition, it is automatically registered from the JSON configuration..

---

## Configuration Priority

The resulting priority order is:

| Priority    | Source                                                           | Format           |
| ----------- | ---------------------------------------------------------------- | ---------------- |
| 1 (highest) | `.pi/settings.json` → `subagents.agentOverrides`                 | JSON             |
| 2           | `~/.pi/agent/settings.json` → `subagents.agentOverrides`         | JSON             |
| 3           | `.pi/agents/*.md`                                                | YAML frontmatter |
| 4           | `~/.pi/agent/agents/*.md`                                        | YAML frontmatter |
| 5 (lowest)  | Built-in agents such as `general-purpose`, `Explore`, and `Plan` | Hardcoded        |

Local project overrides merge on top of global overrides on a per-field basis.

For example, a project configuration can replace only an agent's `model` and `thinking` values while retaining its globally configured `tools` and `systemPrompt`.

---

### Global defaults

`~/.pi/agent/settings.json`

```json
{
  "subagents": {
    "defaultModel": "opencode/deepseek-v4-flash-free",
    "agentOverrides": {
      "Explore": {
        "model": "github-copilot/gpt-5-mini",
        "thinking": "off"
      },
      "reviewer": {
        "model": "opencode/deepseek-v4-flash-free",
        "thinking": "max",
        "tools": ["read", "grep", "find", "ls", "bash"]
      }
    }
  }
}
```

### Project-specific overrides

`.pi/settings.json`

Only the fields that differ from the global configuration need to be specified:

```json
{
  "subagents": {
    "agentOverrides": {
      "reviewer": {
        "model": "openai-codex/gpt-5.4",
        "thinking": "high"
      },
      "fixer": {
        "model": "openai-codex/gpt-5.4",
        "tools": ["read", "grep", "find", "ls", "bash", "edit", "write"],
        "systemPrompt": "You are Fixer — a fast, focused implementation specialist. NO external research, NO delegation, NO planning."
      }
    }
  }
}
```

In this example, `reviewer` inherits its remaining fields from the global configuration.

The `fixer` agent does not need to exist as a built-in or `.md` definition. It is automatically registered from this JSON configuration.

---

## Implementation

* Adds `src/nico-overrides.ts` to read, merge, normalize, and apply JSON overrides
* Exposes `applyNicoOverrides()` from `src/agent-types.ts`
* Applies overrides after normal agent registration in `reloadCustomAgents()`
* Converts compatible fields, including the `skills` representation, into the format expected by `@tintinweb/pi-subagents`
* Leaves existing behavior unchanged when no overrides are configured

This compatibility layer reads `agentOverrides` from Pi's global and project `settings.json` files. It does not add the same option to this extension's own `subagents.json` configuration format.

---

Thanks to [@nicobailon](https://github.com/nicobailon) for creating the original `pi-subagents` configuration format that this compatibility layer builds upon.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
-  "name": "@tintinweb/pi-subagents",
-  "version": "0.14.3",
+  "name": "@esso0428/pi-subagents",
+  "version": "0.15.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
-      "name": "@tintinweb/pi-subagents",
-      "version": "0.14.3",
+      "name": "@esso0428/pi-subagents",
+      "version": "0.15.0",
       "license": "MIT",
       "dependencies": {
         "@sinclair/typebox": "^0.34.49",
@@ -1349,6 +1349,9 @@
       "cpu": [
         "arm64"
       ],
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1365,6 +1368,9 @@
       "integrity": "sha512-AGuJdgKsmJdm4Pych7kv3sqe591ERRaAHW3xjLooiFzn8J+PxUyof++7YZrB5Y5tpnTO+K18Og3taj2NpluCRQ==",
       "cpu": [
         "arm64"
+      ],
+      "libc": [
+        "musl"
       ],
       "license": "MIT",
       "optional": true,
@@ -1383,6 +1389,9 @@
       "cpu": [
         "riscv64"
       ],
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1400,6 +1409,9 @@
       "cpu": [
         "x64"
       ],
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1416,6 +1428,9 @@
       "integrity": "sha512-/DHn+1DrfL6oRaPPWXaOKvonFFrni666fxd+zFqiQEfvBH0tsHVWjq9iqBk0oDp0qaPA72lIMy5BptxISBEhZQ==",
       "cpu": [
         "x64"
+      ],
+      "libc": [
+        "musl"
       ],
       "license": "MIT",
       "optional": true,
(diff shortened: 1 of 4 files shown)


clouatre:
**This would have prevented a real incident for us — details + one subtlety worth documenting.**

pi 0.85.1, @tintinweb/pi-subagents 0.19.0, `defaultProvider: zai`, `OPENROUTER_API_KEY` set, no Anthropic auth. The built-in `Explore` agent pins `anthropic/claude-haiku-4-5`; `resolveModel`'s fuzzy fallback matched the same model under OpenRouter, and 68 subagent sessions silently ran on Claude Haiku via OpenRouter credits (session-log confirmed, reproducible).

We had to work around it with a same-name `.md` — which replaces the whole config, so that's a full 39-line prompt copy plus a generator script, install-time drift check, and weekly CI to keep it aligned with upstream. With this PR, the entire workaround becomes:

```json
{ "subagents": { "agentOverrides": { "Explore": { "model": "zai/glm-5.3-flash", "thinking": "low" } } } }
```

Verified against the patch: keys match the registry exactly (capital-E `"Explore"`; a typo would silently auto-register a new agent rather than error), and only `model`/`thinking`/`systemPrompt`/`disabled`/`tools` are overwritten for existing agents. The PR is against 0.14.3 and ~2 months stale — happy to help test a rebase against 0.19.x.

ESSO0428:
Thanks for the detailed feedback and the great point about agent name case sensitivity! That auto-registration behavior is handy for custom agents, but indeed worth being mindful of for built-ins.

I've just rebased the branch onto 0.19.0. Would love to have your verification with your setup whenever you have time. Let me know if you run into any edge cases!

Issue #185, opened 2026-07-29
Completion notifications for already-joined agents arrive after the final answer


vincelwt:
Background agent completion notifications can be delivered long after they stop being useful — after the final answer, for agents whose results were already read with `get_subagent_result`. Each one costs a turn to answer "already incorporated".

**Why the existing suppression misses it.** The notification is sent with `deliverAs: "followUp"`, which pi delivers only once the agent has no more tool calls. Emitting one mid-run parks it in pi's follow-up queue until the run ends, and a parked message can't be withdrawn. `emitIndividualNudge` does re-check `record.resultConsumed`, but that check runs when the message is enqueued — 200ms after completion (`NUDGE_HOLD_MS`) — while the orchestrator typically joins the agent minutes later. So a 200ms cancellation window is guarding a queue with multi-minute latency.

With pi's default `followUpMode: "one-at-a-time"` the parked notifications then drain one turn each.

**Repro:** spawn a background agent during a long run, join it with `get_subagent_result` a few minutes later, keep working. Its notification still arrives after the final answer.

**From a real session** (pi 0.82.1, extension 0.14.3, macOS):

```
22:09:39  agent spawned (run_in_background: true)
22:10:42  agent completes
22:13:55  get_subagent_result  ->  resultConsumed = true
22:23:40  final answer
22:23:40  notification delivered (13 min late)
22:23:44  extra turn: "Already incorporated; no further action needed."
```

A worse one in the same setup: 7 agents that had finished up to 1h41m earlier were flushed in a 54-second burst after the final answer — 7 wasted turns, and the burst pushed the session into a compaction.

**Suggested fix:** hold due notifications in-process rather than in pi's queue, and flush them at `agent_settled` — the first point where pi won't continue on its own. `agent_end` and `turn_end` both still have retries, compaction, or further tool-calling turns ahead, so emitting there parks them exactly as before. Since each send closure already re-checks `resultConsumed`, deferring the call defers the check, and a joined agent simply never notifies.

I have this implemented with tests and would like to submit it — opening a PR right after this so there's something concrete to react to. Happy to rework it if you'd rather solve it a different way.

Hayston1001:
Real-session reproduction on master (0.19.0) via the **group delivery path** — same root cause as this issue, but on the grouped-notification branch, which I don't think the analysis above covers.

## Repro

[...]

PR #190, opened 2026-07-31
fix: steer background completion notifications

### LINKED ISSUE #188: Deliver background completions at the next safe parent boundary
Background completion notifications should reach an active parent at the next safe steering boundary.

## Problem

Both individual and grouped completions currently use:

```ts
{ deliverAs: "followUp", triggerTurn: true }
```

`triggerTurn` correctly wakes an idle parent, as introduced by #14. During an active run, Pi queues `followUp` until the run would otherwise stop. A parent that keeps making tool calls can therefore miss completed results for minutes and receive them only after its final answer.

Minimal reproduction with Pi 0.82.1 and pi-subagents 0.14.3 at upstream commit `3ed8632`, default smart join mode:

1. Spawn background agents during a tool-heavy parent run.
2. Keep the parent working after the children complete.
3. Observe when terminal records and model-visible notifications appear.

In one session, two terminal records were persisted at 13:52:43 and 13:52:51, but their notifications arrived 131.9s and 163.4s later, after the parent had already retrieved both results and continued working.

This overlaps with #185 but has a different primary requirement:

- #185 / #186 optimize suppression of already-consumed notifications by holding them until `agent_settled`.
- #61 favors a later user-initiated turn so `before_agent_start` hooks run.
- This issue favors exposing an unread completion promptly to a continuously active parent without cancelling an in-flight tool.

Pi documents `steer` as delivery after the current assistant turn finishes its complete tool-call batch and before the next model call. This is also the current behavior of `nicobailon/pi-subagents`: its notifier sets `triggerTurn` and leaves `deliverAs` at Pi's default `steer` mode.

## Proposed behavior

For both completion paths:

```ts
{ deliverAs: "steer", triggerTurn: true }
```

Preserve the existing consumption window and grouping behavior.

Expected behavior:

- Active parent: finish the current complete tool batch normally, then expose the completion before the next model call.
- Idle parent: process the completion without user input.
- Never abort an already-running tool because a child completed.

A notification already accepted by Pi cannot be withdrawn if `get_subagent_result` consumes it later in the same parent turn. That is the trade-off versus #186's settle-time suppression and #61's next-user-turn policy; the maintainer can choose which behavior the extension should guarantee.

Related: #14, #61, #185, #186.

I intend to submit a focused implementation with extension-wiring and real Pi scheduling regression tests.

xz-dev:
## Summary

- deliver individual and grouped background completions through Pi's steering queue
- retain `triggerTurn: true` so idle parents still process completions without user input
- cover extension wiring plus active/idle scheduling through a real headless Pi session

## Behavior

While the parent is active, Pi now exposes a completed child after the current assistant turn's complete tool batch and before the next model call. Already-running tools are not aborted.

This follows the behavior used by `nicobailon/pi-subagents`, whose completion notifier leaves `deliverAs` at Pi's default `steer` mode while setting `triggerTurn`.

There is an explicit policy trade-off with the related proposals:

- #185 / #186 prioritize suppressing notifications for results consumed before delivery by holding them until `agent_settled`.
- #61 prioritizes a later user-initiated turn so `before_agent_start` hooks run.
- This PR prioritizes prompt delivery to a continuously active parent. Once Pi has accepted a steering message, a later same-turn `get_subagent_result` cannot retract it.

The maintainer can choose which guarantee the extension should provide; this PR keeps the alternatives separate rather than combining their lifecycle models.

Closes #188.

Related: #14, #61, #185, #186.

## Tests

- `npx vitest run test/completion-notification-wiring.test.ts test/steer-scheduling-e2e.test.ts`
- `npm run lint`
- `npm run typecheck`
- `npm run test` — 759 passed, 5 skipped
- `npm run build`
- `git diff --check upstream/master...HEAD`

The active-parent test waits for the public `subagents:completed` event while a controlled parent tool remains blocked, verifies the notification is not injected during that tool batch and the tool is not aborted, then verifies the immediately following parent model call sees the completion.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -327,7 +327,7 @@ export default function (pi: ExtensionAPI) {
       content: notification + footer,
       display: true,
       details: buildNotificationDetails(record, 500, agentActivity.get(record.id)),
-    }, { deliverAs: "followUp", triggerTurn: true });
+    }, { deliverAs: "steer", triggerTurn: true });
   }
 
   function sendIndividualNudge(record: AgentRecord) {
@@ -365,7 +365,7 @@ export default function (pi: ExtensionAPI) {
           content: `Background agent group completed: ${label}\n\n${notifications}\n\nUse get_subagent_result for full output.`,
           display: true,
           details,
-        }, { deliverAs: "followUp", triggerTurn: true });
+        }, { deliverAs: "steer", triggerTurn: true });
       });
       widget.update();
     },
--- a/test/completion-notification-wiring.test.ts
+++ b/test/completion-notification-wiring.test.ts
@@ -1,0 +1,161 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import type * as AgentRunnerModule from "../src/agent-runner.js";
+
+vi.mock("../src/agent-runner.js", async () => {
+  const actual = await vi.importActual<typeof AgentRunnerModule>("../src/agent-runner.js");
+  return {
+    ...actual,
+    runAgent: vi.fn(),
+  };
+});
+
+import { runAgent } from "../src/agent-runner.js";
+import subagentsExtension from "../src/index.js";
+
+interface RegisteredTool {
+  name: string;
+  execute: (
+    toolCallId: string,
+    params: Record<string, unknown>,
+    signal: AbortSignal | undefined,
+    onUpdate: undefined,
+    ctx: ReturnType<typeof makeHeadlessCtx>,
+  ) => Promise<unknown>;
+}
+
+function makePi() {
+  const tools = new Map<string, RegisteredTool>();
+  const handlers = new Map<string, (...args: never[]) => unknown>();
+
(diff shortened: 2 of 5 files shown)


xz-dev:
PR #190 and #186 address the same delayed-completion symptom, but they intentionally choose different delivery boundaries.

#190 delivers with `steer` at Pi’s next safe steering boundary. It does not cancel an active tool batch, but it lets an active parent see a newly completed subagent result before the parent would otherwise settle, while still waking an idle parent.

#186 parks completion while the parent is busy and delivers after `agent_settled`. That creates a later `resultConsumed` re-check window and can suppress a notification consumed during the active turn, but it also delays a work-relevant completion until that turn has already settled.

The fork currently retains #190’s next-safe-boundary behavior because active-parent timeliness is the requirement diagnosed in #188. It would help to make the boundary choice explicit when deciding between these PRs; they are alternatives rather than additive fixes.

PR #201, opened 2026-08-04
feat(isolation): add 'off' option to explicitly disable isolation

Munsen:
GPT 5.6 always uses `worktree` as the `isolation` when calling an `Agent`.
It was useless even when I was asked to leave the `isolation` empty. 

After adding the `off` option to isolation, GPT 5.6 did not use `worktree` in isolation even without any instruction.

### DIFF
--- a/src/custom-agents.ts
+++ b/src/custom-agents.ts
@@ -80,7 +80,7 @@ function loadFromDir(dir: string, agents: Map<string, AgentConfig>, source: "pro
       runInBackground: fm.run_in_background != null ? fm.run_in_background === true : undefined,
       isolated: fm.isolated != null ? fm.isolated === true : undefined,
       memory: parseMemory(fm.memory),
-      isolation: fm.isolation === "worktree" ? "worktree" : undefined,
+      isolation: (fm.isolation === "worktree" || fm.isolation === "off") ? fm.isolation : undefined,
       enabled: fm.enabled !== false,  // default true; explicitly false disables
       source,
     });
--- a/src/index.ts
+++ b/src/index.ts
@@ -803,7 +803,7 @@ Notes:
 - Parallel work: one message, multiple Agent calls, run_in_background: true on each. You are notified when background agents finish — never poll or sleep.
 - The result is not shown to the user — summarize it for them. Verify an agent's claimed code changes before reporting work done.
 - resume continues a previous agent by ID; steer_subagent messages a running one.
-- isolation: "worktree" runs the agent in an isolated git worktree; changes land on a branch.`;
+- isolation: "worktree" runs the agent in an isolated git worktree; changes land on a branch. isolation: "off" explicitly disables isolation (same as omitting).`;
 
   const fullAgentToolDescription = `Launch a new agent to handle complex, multi-step tasks autonomously. Each agent type has specific capabilities and tools available to it.
 
@@ -954,9 +954,14 @@ Terse command-style prompts produce shallow, generic work.
         }),
       ),
       isolation: Type.Optional(
-        Type.Literal("worktree", {
-          description: 'Set to "worktree" to run the agent in a temporary git worktree (isolated copy of the repo). Changes are saved to a branch on completion.',
-        }),
+        Type.Union([
+          Type.Literal("worktree", {
+            description: 'Set to "worktree" to run the agent in a temporary git worktree (isolated copy of the repo). Changes are saved to a branch on completion.',
+          }),
+          Type.Literal("off", {
+            description: 'Explicitly disable isolation. Same effect as omitting the field.',
+          }),
+        ]),
       ),
       ...scheduleParam,
     }),
--- a/src/types.ts
+++ b/src/types.ts
@@ -18,7 +18,7 @@ export const DEFAULT_AGENT_NAMES = ["general-purpose", "Explore", "Plan"] as con
 export type MemoryScope = "user" | "project" | "local";
 
 /** Isolation mode for agent execution. */
-export type IsolationMode = "worktree";
+export type IsolationMode = "worktree" | "off";
 
 /** Unified agent configuration — used for both default and user-defined agents. */
 export interface AgentConfig {
--- a/test/custom-agents.test.ts
+++ b/test/custom-agents.test.ts
@@ -673,6 +673,18 @@ Isolated.`);
     expect(result.get("isolated-wt")!.isolation).toBe("worktree");
   });
 
+  it("parses isolation: off (explicit opt-out)", () => {
+    writeAgent("isolated-off", `---
+description: Explicit off
+isolation: off
+---
+
+Off.`);
+
+    const result = loadCustomAgents(tmpDir);
+    expect(result.get("isolated-off")!.isolation).toBe("off");
+  });
+
   it("isolation defaults to undefined when omitted", () => {
     writeAgent("no-isolation", `---
 description: Normal
--- a/test/tool-description-mode.test.ts
+++ b/test/tool-description-mode.test.ts
@@ -122,6 +122,7 @@ describe("toolDescriptionMode", () => {
       "resume",
       "steer_subagent",
       'isolation: "worktree"',
+      'isolation: "off"',
       ".pi/agents/",
       "self-contained",
     ]) {


tintinweb:
can you provide a session log? would want to avoid adding to context unless necessary.

Munsen:
Is it correct that the session log refers to the tool section?
```
{
      "type": "function",
      "name": "Agent",
      "description": "Launch a new agent to handle complex, multi-step tasks autonomously. Each agent type has specific capabilities and tools available to it.\n\nAvailable agent types and the tools they have access to:\n- general-purpose: General-purpose agent for researching complex questions, searching for code, and executing multi-step tasks. When you are searching for a keyword or file and are not confident that you will find the right match in the first few tries use this agent to perform the search for you. (deepseek-v4-flash) (Tools: *)\n- claim-verifier: Adversarial finding verifier. Grounds each supplied claim against actual repository state and emits one `FINDING <id> | <tag> | <justification>` row per input, with tags Verified / Weakened / Falsified. Tier: git-analyzer (+ `bash` for `git show`). Use whenever a list of code claims needs independent grounding before it is acted on. (minimax-m3) (Tools: read, grep, find, ls, bash)\n- code-reviewer: Performs comprehensive code reviews, pull request evaluations, and quality checks. Invoke when asked to review code quality, correctness, performance, security, or maintainability. (code-reviewer) (Tools: read, grep, find, ls)\n- codebase-analyzer: Analyzes codebase implementation details. Call the codebase-analyzer agent when you need to find detailed information about specific components. As always, the more detailed your request prompt, the better! :) (minimax-m3) (Tools: read, grep, find, ls)\n- codebase-locator: Locates files, directories, and components relevant to a feature or task. Call `codebase-locator` with a human-language prompt describing what you're looking for. A \"super grep/find/ls\" tool. Reach for it when you would otherwise reach for grep, find, or ls more than once. (deepseek-v4-flash) (Tools: grep, find, ls)\n- codebase-pattern-finder: codebase-pattern-finder is a useful subagent_type for finding similar implementations, usage examples, or existing patterns that can be modeled after. It will give you concrete code examples based on what you're looking for! It's sorta like codebase-locator, but it will not only tell you the location of files, it will also give you code details! (minimax-m3) (Tools: grep, find, read, ls)\n- diff-auditor: Row-only patch auditor. Walks a patch against a caller-supplied surface-list and emits one pipe-delimited row per finding (`file:line | verbatim | surface-id | note`). Use whenever a diff needs evidence-only enumeration of matching patterns, with no narrative or severity. (minimax-m3) (Tools: read, grep, find, ls)\n- git-master: Git operations specialist handling atomic commits, rebases, history search, and all git workflows. Triggers on 'commit', 'push', 'rebase', 'merge', 'blame', 'bisect', etc. Delegate ALL git tasks to this agent. (deepseek-v4-flash) (Tools: read, grep, find, ls, bash)\n- integration-scanner: Finds what connects to a given component or area: inbound references, outbound dependencies, config registrations, event subscriptions. The reverse-reference counterpart to codebase-locator. Use when you need to understand what calls, depends on, or wires into a component. (minimax-m3) (Tools: grep, find, ls)\n- package-scout: Searches and evaluates external packages — summarizes downloads, stars, last update, license, and compatibility for Plan Agent reference. Invoked on-demand during planning when package selection is needed. (deepseek-v4-flash) (Tools: read, grep, find, ls, bash)\n- peer-comparator: Pairwise peer-invariant comparator. Given `(new_file, peer_file)` pairs, tags each peer invariant Mirrored / Missing / Diverged / Intentionally-absent against the new file. Use when an entity parallels an existing sibling (aggregate, service, handler, reducer, repository) and the new file must be checked against the peer's public surface. (minimax-m3) (Tools: read, grep, find, ls)\n- web-search-researcher: Do you find yourself desiring information that you don't quite feel well-trained (confident) on? Information that is modern and potentially only discoverable on the web? Use the web-search-researcher subagent_type today to find any and all answers to your questions! It will research deeply to figure out and attempt to answer your questions! If you aren't immediately satisfied you can get your money back! (Not really - but you can re-run web-search-researcher with an altered prompt in the event you're not satisfied the first time) (minimax-m3) (Tools: read, grep, find, ls)\n\nCustom agents can be defined in .pi/agents/<name>.md (project) or /home/briancha/.pi/agent/agents/<name>.md (global) — they are picked up automatically. Project-level agents override global ones. Creating a .md file with the same name as a default agent overrides it.\n\nWhen using the Agent tool, specify a subagent_type parameter to select which agent type to use.\n\n## When not to use\n\nIf the target is already known, use a direct tool — `read` for a known path, `grep`/`find` for a specific symbol or string. Reserve this tool for open-ended questions that span the codebase, or tasks that match an available agent type.\n\n## Usage notes\n\n- Always include a short (3-5 word) description summarizing what the agent will do (shown in UI).\n- When you launch multiple agents for independent work, send them in a single message with multiple tool uses, with run_in_background: true on each, so they run concurrently. If the user specifies that they want agents run \"in parallel\", you MUST send a single message with multiple tool calls. Foreground calls run sequentially — only one executes at a time.\n- When the agent is done, it returns a single message back to you. The result is not visible to the user — to show the user, send a text message with a concise summary.\n- Trust but verify: an agent's summary describes what it intended to do, not necessarily what it did. When an agent writes or edits code, check the actual changes before reporting work as done.\n- Use run_in_background for work you don't need immediately. You will be notified when it completes — do NOT poll or sleep waiting for it. Continue with other work or respond to the user instead.\n- Foreground vs background: use foreground (default) when you need the agent's results before you can proceed. Use background when you have genuinely independent work to do in parallel.\n- Use resume with an agent ID to continue a previous agent's work. A new (non-resume) Agent call starts a fresh agent with no memory of prior runs, so the prompt must be self-contained.\n- Use steer_subagent to send mid-run messages to a running background agent.\n- Clearly tell the agent whether you expect it to write code or just to do research (search, file reads, etc.), since it is not aware of the user's intent.\n- If an agent's description says it should be used proactively, try to use it without the user having to ask for it first.\n- Use model to specify a different model (as \"provider/modelId\", or fuzzy e.g. \"haiku\", \"sonnet\").\n- Use thinking to control extended thinking level.\n- Use inherit_context if the agent needs the parent conversation history.\n- Use isolation: \"worktree\" to run the agent in an isolated git worktree (safe parallel file modifications). The worktree is automatically cleaned up if the agent makes no changes; otherwise the path and branch are returned in the result.\n- Use `schedule` only when the user explicitly asked for scheduled / recurring / delayed execution (e.g. \"every Monday\", \"in an hour\"). Don't auto-schedule from vague intent like \"monitor X\" — run once now or ask.\n\n## Writing the prompt\n\nProvide clear, detailed prompts so the agent can work autonomously. Brief it like a smart colleague who just walked into the room — it hasn't seen this conversation, doesn't know what you've tried, doesn't understand why this task matters.\n- Explain what you're trying to accomplish and why.\n- Describe what you've already learned or ruled out.\n- Give enough context about the surrounding problem that the agent can make judgment calls rather than just following a narrow instruction.\n- If you need a short response, say so (\"report in under 200 words\").\n- Lookups: hand over the exact command. Investigations: hand over the question — prescribed steps become dead weight when the premise is wrong.\n\nTerse command-style prompts produce shallow, generic work.\n\n**Never delegate understanding.** Don't write \"based on your findings, fix the bug\" or \"based on the research, implement it.\" Those phrases push synthesis onto the agent instead of doing it yourself. Write prompts that prove you understood: include file paths, line numbers, what specifically to change.",
      "parameters": {
        "type": "object",
        "required": [
          "prompt",
          "description",
          "subagent_type"
        ],
        "properties": {
          "prompt": {
            "type": "string",
            "description": "The task for the agent to perform."
          },
          "description": {
            "type": "string",
            "description": "A short (3-5 word) description of the task (shown in UI)."
          },
          "subagent_type": {
            "type": "string",
            "description": "The type of specialized agent to use. Available types: general-purpose, claim-verifier, code-reviewer, codebase-analyzer, codebase-locator, codebase-pattern-finder, diff-auditor, git-master, integration-scanner, package-scout, peer-comparator, web-search-researcher. Custom agents from .pi/agents/*.md (project) or /home/briancha/.pi/agent/agents/*.md (global) are also available."
          },
          "model": {
            "type": "string",
            "description": "Optional model override. Accepts \"provider/modelId\" or fuzzy name (e.g. \"haiku\", \"sonnet\"). Omit to use the agent type's default."
          },
          "thinking": {
            "type": "string",
            "description": "Thinking level: off, minimal, low, medium, high, xhigh, max. Overrides agent default."
          },
          "max_turns": {
            "type": "number",
            "description": "Maximum number of agentic turns before stopping. Omit for unlimited (default).",
            "minimum": 1
          },
          "run_in_background": {
            "type": "boolean",
            "description": "Set to true to run in background. Returns agent ID immediately. You will be notified on completion."
          },
          "resume": {
            "type": "string",
            "description": "Optional agent ID to resume from. Continues from previous context."
          },
          "isolated": {
            "type": "boolean",
            "description": "If true, agent gets no extension/MCP tools — only built-in tools."
          },
          "inherit_context": {
            "type": "boolean",
            "description": "If true, fork parent conversation into the agent. Default: false (fresh context)."
          },
          "isolation": {
            "anyOf": [
              {
                "type": "string",
                "const": "worktree",
                "description": "Set to \"worktree\" to run the agent in a temporary git worktree (isolated copy of the repo). Changes are saved to a branch on completion."
              },
              {
                "type": "string",
                "const": "off",
                "description": "Explicitly disable isolation. Same effect as omitting the field."
              }
            ]
          },
          "schedule": {
            "type": "string",
            "description": "Opt-in only — fire later instead of now. Omit to run immediately (the default, almost always correct). Formats: 6-field cron (\"0 0 9 * * 1\" = 9am Mon), interval (\"5m\"/\"1h\"), one-shot (\"+10m\" or ISO). Forces run_in_background; incompatible with inherit_context and resume. Returns job ID."
          }
        }
      },
      "strict": false
    }
```

tintinweb:
Can you provide a session transcript that shows how gpt chose to add worktree:isolated while it shouldnt?

Munsen:
It is no longer reproduced. Close PR

PR #202, opened 2026-08-04
feat(usage): report subagent cost on tool results via ToolResultMessage.usage

### LINKED ISSUE #203: Subagent spend is missing from the parent session's cost
## Problem

Subagents run in their own `AgentSession`. Their LLM spend never reaches the parent session's cost accounting. Every surface that reads parent session cost under-reports total spend. This includes pi core's `getUsageCostBreakdown`, `/session`, and statusline extensions such as pi-starship and pi-powerline-footer. In a test on pi 0.83.0, two one-turn subagent runs spent $0.066 and none of it reached the parent session's cost accounting.

## Proposal

Pi provides a canonical channel for tool spend: `ToolResultMessage.usage`. Pi core aggregates that field into its "Tools/summaries" cost bucket, and statusline extensions sum the same field. The extension already accumulates per-agent token usage from `message_end` events. It does not track cost and does not attach anything to the tool results it returns.

I propose three changes:

1. Extend the lifetime usage accumulator with `cacheRead` and `cost`.
2. Attach the not-yet-reported usage slice to tool results at terminal points: foreground `Agent` return, `resume` return, and terminal `get_subagent_result`.
3. Advance a per-record watermark on each report so resumes and repeated result fetches never report the same spend twice.

I implemented and tested this in #202.

unrelentingfox:
Closes #203.

## Problem

Subagent LLM spend is invisible to the parent session's cost accounting. Subagents run in their own `AgentSession`, and the `Agent` and `get_subagent_result` tool results carry token counts as display text only. Every surface that reads parent session cost under-reports total spend: pi core's `getUsageCostBreakdown`, `/session`, and statusline extensions such as pi-starship and pi-powerline-footer. In a test on pi 0.83.0, two one-turn subagent runs spent $0.066 and none of it reached the parent session's cost accounting.

## Solution

Pi provides a channel for exactly this: `ToolResultMessage.usage` ("Usage from the tool execution itself"). Pi core aggregates it into the "Tools/summaries" cost bucket, and statusline extensions sum the same field. This PR attaches the subagent's accumulated usage, including `cost.total`, to the tool results the extension returns.

### Changes

- `LifetimeUsage` gains `cacheRead` and `cost`. The token display total still excludes `cacheRead` (#38 unchanged). Summing `cacheRead` across `message_end` events is billing-correct for the usage report because each call re-reads and is billed for the cached prefix.
- The `message_end` handlers in `runAgent` and `resumeAgent` forward `cacheRead` and `usage.cost.total`.
- `takeUnreportedUsage()` returns the not-yet-reported slice as a pi-ai `Usage` and advances a per-record watermark (`AgentRecord.reportedUsage`). Delta semantics make repeated reporting safe. A resume reports only new spend. A second `get_subagent_result` reports nothing. The parent session never double-counts.
- Usage attaches on foreground `Agent` completion and failure, `resume` returns, and terminal `get_subagent_result`. Background spawn returns and still-running polls attach nothing.
- Nested subagents keep propagating usage up ancestor `lifetimeUsage`, now including cost, so one top-level tool result covers the whole subtree.

## Testing

All four checks pass locally: `lint` (clean), `typecheck`, `test` (824 passed, 5 skipped), and `build`. `test:e2e` also passes (49 passed, 5 skipped).

New `takeUnreportedUsage` unit tests cover: full report on first call, `undefined` on empty report, delta-only report after resume, no double count on repeat calls, and negative-delta clamp.

Verified end-to-end on pi 0.83.0 with bedrock:

| Path | `usage` on toolResult |
|---|---|
| Failed spawn (model not in scope) | none |
| Foreground `Agent` completion | full usage, cost $0.057 |
| Background spawn return | none (still running) |
| 1st `get_subagent_result` | full usage, cost $0.009 |
| 2nd `get_subagent_result` | none (already reported) |

Statusline extensions such as pi-starship sum the same toolResult field, so they pick up subagent cost with no changes.

### DIFF
--- a/src/agent-manager.ts
+++ b/src/agent-manager.ts
@@ -14,7 +14,7 @@ import type { Model } from "@earendil-works/pi-ai";
 import type { AgentSession, ExtensionAPI, ExtensionContext } from "@earendil-works/pi-coding-agent";
 import { resumeAgent, runAgent, type ToolActivity } from "./agent-runner.js";
 import type { AgentInvocation, AgentRecord, IsolationMode, SubagentType, ThinkingLevel } from "./types.js";
-import { addUsage } from "./usage.js";
+import { addUsage, emptyUsage, type LifetimeUsage } from "./usage.js";
 import { cleanupWorktree, createWorktree, pruneWorktrees, } from "./worktree.js";
 
 export type OnAgentComplete = (record: AgentRecord) => void;
@@ -106,7 +106,7 @@ interface SpawnOptions {
   /** Called at the end of each agentic turn with the cumulative count. */
   onTurnEnd?: (turnCount: number) => void;
   /** Called once per assistant message_end with that message's usage delta. */
-  onAssistantUsage?: (usage: { input: number; output: number; cacheWrite: number }) => void;
+  onAssistantUsage?: (usage: LifetimeUsage) => void;
   /** Called when the session successfully compacts. */
   onCompaction?: (info: CompactionInfo) => void;
   /** Nesting depth: top-level subagent = 1. */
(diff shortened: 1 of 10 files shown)


tintinweb:
superseded by #243 - enabled `reportCost` + `showCost` off by default. scheduled for v0.18.0.

PR #214, opened 2026-08-09
Support run_in_background on resume

akram-ahrardi:
## Problem

Resuming an agent silently ignores `run_in_background`. In the `Agent` tool handler the `if (params.resume)` branch returns before the `if (runInBackground)` branch, and `AgentManager.resume()` only ever `await`s the run inline — so a resumed agent always blocks the main loop until it finishes, unlike a fresh background spawn. There's no error; the flag is just ignored. (Confirmed on `master` @ 0.14.3.)

## Change

- `AgentManager.resume(id, prompt, signal?, options?)` gains an `isBackground` option. When set, the run is detached via a new private `startResume()` that mirrors `startAgent`'s background path:
  - fresh `AbortController`, so `/agents` stop and steering target this run — and the run isn't tied to the tool-call signal, which resolves the moment the tool returns;
  - concurrency-pool accounting + queueing at `maxConcurrent`;
  - output-file flush, owned-child cleanup, and `onComplete` (so the normal completion notification fires);
  - returns immediately with the record still `running` (or `queued`) and sets `record.promise`, so `get_subagent_result(wait: true)` works on it.
- The queue entry was generalized from `{ id, args }` to `{ id, start }` so a resume can be queued alongside spawns.
- The `Agent` tool gains a background-resume branch: returns the agent ID immediately, attaches a fresh transcript and streams it on the already-existing session, emits `subagents:created`, and registers the run in the widget/fleet — matching the background-spawn path.
- Foreground resume behavior is unchanged.

## Tests

Added 5 unit tests in `test/agent-manager.test.ts`: background resume returns immediately + notifies on completion; a failed final turn maps to `error` and still notifies; activity/usage callbacks are forwarded; a resume queues when the concurrency pool is full; foreground resume is unchanged.

`npm run lint`, `npm run typecheck`, `npm run build`, and `npm run test` (830 tests) all pass. README updated per CONTRIBUTING; CHANGELOG left untouched per CONTRIBUTING.


tintinweb:
included your work in #226; your commits are credited there. thanks!

PR #220, opened 2026-08-11
Expose cacheRead/cacheWrite in usage payload + fix prompt cache prefix ordering

mateominato:
## What this does

**`src/usage.ts`** — the emitted usage payload had no cache-read/cache-write fields at all, so downstream telemetry/observability tooling couldn't see cache activity for a subagent dispatch. This adds them additively:

- `cacheRead` is optional (`cacheRead?: number`) and last-write-wins — each turn overwrites rather than sums, since upstream reports it as a cumulative per-turn snapshot, not a delta.
- `getLifetimeTotal()` is unchanged; this does **not** reintroduce the double-counting issue from #38 — `cacheRead` is deliberately never summed, only overwritten with the latest observed value each turn.

**`src/prompts.ts`** — reorders `buildAgentPrompt` so the static per-agent-type content is emitted before the variable env block, restoring a stable, cacheable prompt prefix across spawns. Wording is unchanged, only ordering, applied to both `replace` and `append` modes.

These are two independent changes bundled into one branch for convenience. Happy to split into two PRs if that's easier to review.

Tests, build, and lint all pass.

Ref: #38

### DIFF
--- a/src/agent-manager.ts
+++ b/src/agent-manager.ts
@@ -106,7 +106,7 @@ interface SpawnOptions {
   /** Called at the end of each agentic turn with the cumulative count. */
   onTurnEnd?: (turnCount: number) => void;
   /** Called once per assistant message_end with that message's usage delta. */
-  onAssistantUsage?: (usage: { input: number; output: number; cacheWrite: number }) => void;
+  onAssistantUsage?: (usage: { input: number; output: number; cacheWrite: number; cacheRead: number }) => void;
   /** Called when the session successfully compacts. */
   onCompaction?: (info: CompactionInfo) => void;
   /** Nesting depth: top-level subagent = 1. */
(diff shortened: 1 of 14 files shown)


mateominato:
Hey, thanks for this package — closing our PR since we've stopped consuming it internally, so we won't be able to maintain the change going forward.

One thing worth flagging from digging into #38: Anthropic documents `cache_read_input_tokens` as tokens retrieved from the cache *for this request* — a per-call figure, not cumulative. That matches other clients (e.g. `pi-ai` assigns it fresh per response rather than accumulating), and it's why upstream keeps two readings: one summing `cacheRead` across turns for billing totals, one using only the latest turn for context size.

If `cacheRead` is genuinely per-call here, last-write-wins (as our PR did) would likely under-report cache spend across multi-turn sessions — summing for cost while keeping a latest-turn figure for context size may be worth a look. Could be we're missing context, so just an observation. Thanks again for the project!

PR #228, opened 2026-08-14
fix: fire session_shutdown before disposing child sessions

v-kuskov:
## Problem

Disposing an in-process child `AgentSession` can crash the whole pi process with
an uncaught `This extension ctx is stale after session replacement or reload`.
Reproduced with `@4fu/pi-pwsh`: it keeps a session-scoped heartbeat that touches
its ctx, and the throw lands after the subagent is done with the session, so
nothing catches it.

## Root cause

pi's `AgentSessionRuntime.dispose()` emits `session_shutdown` before invalidating
the extension runner; plain `AgentSession.dispose()` does not. Subagents build
child sessions via `createAgentSession` directly, so their extensions never get
the stop event and keep touching a stale ctx.

## Fix

Emit `session_shutdown` (reason `quit`) on the child's extension runner before
disposing, mirroring `AgentSessionRuntime.dispose()`:

- await the emit so shutdown handlers finish (stop heartbeats) before the runner
  is invalidated
- detach the session from the record before the async emit so a concurrent
  remove/dispose can't double-fire
- swallow handler/dispose errors so they can't block disposal or escape as
  unhandled rejections

## Caveat

This relies on the undocumented `session.extensionRunner.emit` seam. The durable
fix is pi-core exposing `AgentSession.shutdown()`; tracked separately.

## Tests

`test/agent-manager.test.ts` — "fires session_shutdown before disposing a
session with extensions" pins the `shutdown` → `dispose` ordering.

### DIFF
--- a/src/agent-manager.ts
+++ b/src/agent-manager.ts
@@ -147,6 +147,41 @@ interface ResumeOptions {
   onStarted?: () => void;
 }
 
+/**
+ * Fire session_shutdown on a child session's extension runner, then dispose it.
+ *
+ * WHY (hack, not a pi-core fix): pi's `AgentSession.dispose()` invalidates the
+ * extension runner WITHOUT emitting `session_shutdown` — the lifecycle stop
+ * event extensions use to close session-scoped resources (heartbeats, open
+ * connections). `AgentSessionRuntime.dispose()` emits it before disposing;
+ * `AgentSession.dispose()` does not. Extensions that keep touching their ctx
+ * after invalidation throw "This extension ctx is stale after session
+ * replacement or reload", surfacing as an uncaughtException that can take the
+ * whole process down (seen with @4fu/pi-pwsh). Emitting the event ourselves via
+ * the public `session.extensionRunner` getter reproduces what
+ * `AgentSessionRuntime.dispose()` does, without a pi-core change.
+ *
+ * Order matters: the emit must resolve BEFORE dispose(), so a shutdown handler
+ * finishes (stopping its heartbeat) before the runner is invalidated. A handler
+ * that hangs blocks disposal — the same behavior as AgentSessionRuntime.dispose().
+ */
+async function disposeSession(session: AgentSession | undefined): Promise<void> {
+  if (!session) return;
+  const runner = session.extensionRunner;
+  if (runner) {
+    try {
+      await runner.emit({ type: "session_shutdown", reason: "quit" });
+    } catch {
+      // A throwing handler must not block disposal or crash the parent.
+    }
+  }
+  try {
+    session.dispose();
+  } catch {
+    // dispose() is best-effort cleanup.
+  }
+}
+
 export class AgentManager {
   private agents = new Map<string, AgentRecord>();
   private cleanupInterval: ReturnType<typeof setInterval>;
@@ -759,11 +794,13 @@ export class AgentManager {
     return true;
   }
 
-  /** Dispose a record's session and remove it from the map. */
+  /** Dispose a record's session (firing session_shutdown first) and remove it from the map. */
   private removeRecord(id: string, record: AgentRecord): void {
-    record.session?.dispose?.();
+    const session = record.session;
+    // Eager detach: a second removeRecord/dispose must not re-emit on this session.
     record.session = undefined;
     this.agents.delete(id);
+    void disposeSession(session);
   }
 
   private cleanup() {
@@ -840,10 +877,13 @@ export class AgentManager {
     clearInterval(this.cleanupInterval);
     // Clear queue
     this.queue = [];
-    for (const record of this.agents.values()) {
-      record.session?.dispose();
-    }
+    const sessions = [...this.agents.values()]
+      .map(record => record.session)
+      .filter((session): session is AgentSession => session !== undefined);
+    // Eager detach before the async emit so a concurrent removeRecord can't re-emit.
+    for (const record of this.agents.values()) record.session = undefined;
     this.agents.clear();
+    void Promise.all(sessions.map(disposeSession));
     // Prune any orphaned git worktrees (crash recovery)
     try { pruneWorktrees(process.cwd()); } catch { /* ignore */ }
     // Also prune repos that caller-supplied cwds created worktrees in — a clean
--- a/test/agent-manager.test.ts
+++ b/test/agent-manager.test.ts
@@ -484,6 +484,37 @@ describe("AgentManager — Bug 3 clearCompleted", () => {
     expect(disposeSpy).toHaveBeenCalledOnce();
   });
 
+  it("fires session_shutdown before disposing a session with extensions", async () => {
+    manager = new AgentManager();
+    const order: string[] = [];
+    const sess = {
+      extensionRunner: {
+        emit: vi.fn(async () => {
+          order.push("shutdown");
+        }),
+      },
+      dispose: vi.fn(() => {
+        order.push("dispose");
+      }),
+    };
+    vi.mocked(runAgent).mockResolvedValue({
+      responseText: "done",
+      session: sess as any,
+      aborted: false,
+      steered: false,
+    });
+
+    const id = manager.spawn(mockPi, mockCtx, "general-purpose", "test", {
+      description: "test",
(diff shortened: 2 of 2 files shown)


tintinweb:
Concerns below, mostly in the sync→async conversion. Concretely, what I'd like before this merges:

  1. **Required:** make `dispose()` return its promise and `await` it at the call site — without this the motivating case isn't actually fixed on quit (detail in §1).
  2. **Required:** bound the emit with a timeout before falling through to `session.dispose()` (§2).
  3. **Nice to have:** the `hasHandlers` guard, the `try` placement, and a doc-comment correction (§3).
  4. **Question, not blocking:** whether the runtime layer was considered instead of a replica (§4). A note in the code is enough if the replica is the deliberate choice.

  ---

  **1. Delivery on the quit path is tick-dependent, and pi-pwsh is on the losing side of it.**

  [`AgentManager.dispose()`](https://github.com/tintinweb/pi-subagents/blob/0c88e522a79f405ded94f875f868fe533e138ec5/src/agent-manager.ts#L876) stays synchronous while its work becomes [`void Promise.all(sessions.map(disposeSession))`](https://github.com/tintinweb/pi-subagents/blob/0c88e522a79f405ded94f875f868fe533e138ec5/src/agent-manager.ts#L886). Its only caller is [`manager.dispose()` in the `session_shutdown` handler](https://github.com/tintinweb/pi-subagents/blob/0c88e522a79f405ded94f875f868fe533e138ec5/src/index.ts#L652), which pi awaits inside `runtimeHost.dispose()` before [`process.exit(0)`](https://github.com/earendil-works/pi/blob/v0.80.6/packages/coding-agent/src/modes/interactive/interactive-mode.ts#L3528-L3535), with only synchronous work in between.

  I replicated that path against `AgentManager` and measured state at the `process.exit(0)` point, varying how much async work the child's `session_shutdown` handler does:

  ```
  handler settles in ≤4 microtask ticks   -> emit + session.dispose() both complete
  handler settles in 5 microtask ticks    -> session.dispose() dropped
  handler settles in 6+ ticks             -> emit dropped too
  handler awaits any timer / I/O          -> both dropped
  ```

  @4fu/pi-pwsh lands past that budget. Its [handler](https://github.com/4fuu/pi-pwsh/blob/aa64ed877fbe0c657345974fc37bfce03c517242/src/index.ts#L355-L366) awaits `notifications.close()` and then [`sessions.close()`](https://github.com/4fuu/pi-pwsh/blob/aa64ed877fbe0c657345974fc37bfce03c517242/src/session-runtime.ts#L107-L112), which awaits `rpc.stop()` and `ptys.close()` — real I/O. On quit, only the synchronous head of that handler runs: [`clearInterval`](https://github.com/4fuu/pi-pwsh/blob/aa64ed877fbe0c657345974fc37bfce03c517242/src/task-notifications.ts#L61-L67) does stop the heartbeat, but RPC and PTY teardown are dropped, and so is `session.dispose()` — meaning [`agent.abort()` and `cleanupSessionResources(sessionId)`](https://github.com/earendil-works/pi/blob/v0.80.6/packages/coding-agent/src/core/agent-session.ts#L799-L816), which the previous synchronous `record.session?.dispose()` always ran. The `removeRecord` sweep path has no exit pressure and does complete; it's quit specifically that's short.

  [`src/index.ts:636`](https://github.com/tintinweb/pi-subagents/blob/0c88e522a79f405ded94f875f868fe533e138ec5/src/index.ts#L636) is already `async`, so:

  ```ts
  // agent-manager.ts
  dispose(): Promise<void> {
    // ...
    const disposed = Promise.all(sessions.map(disposeSession)).then(() => {});
    // ... worktree pruning ...
    return disposed;
  }

  // index.ts:652

// index.ts:652
await manager.dispose();
```

With that, a handler awaiting a 20ms timer yields `["child-shutdown", "child-dispose", "parent-dispose"]` before exit. `tsc --noEmit` stays clean and no other call site needs to change.

One note for whatever test covers this: the new test's `vi.waitFor` waits out exactly the deferral the quit path doesn't have, so it passes either way. Asserting on state at a simulated `process.exit(0)` — after an awaited handler chain, with a handler that awaits a timer — is what actually pins the behavior. Same reason this may look fine in manual testing: on SIGTERM/SIGHUP pi awaits [`drainInput(1000)` *after* `runtimeHost.dispose()`](https://github.com/earendil-works/pi/blob/v0.80.6/packages/coding-agent/src/modes/interactive/interactive-mode.ts#L3504-L3516), so only Ctrl+D and `/quit` are tight.

**2. A hanging handler leaks the session permanently.**

[`removeRecord()`](https://github.com/tintinweb/pi-subagents/blob/0c88e522a79f405ded94f875f868fe533e138ec5/src/agent-manager.ts#L798-L803) detaches `record.session` and drops the map entry before `void disposeSession(session)`, so the detached promise holds the only remaining reference, and [`runner.emit`](https://github.com/earendil-works/pi/blob/v0.80.6/packages/coding-agent/src/core/extensions/runner.ts#L749-L767) awaits each handler with no timeout.

Confirmed with a handler that never settles: after `clearCompleted()` — the same path the [60s sweep](https://github.com/tintinweb/pi-subagents/blob/0c88e522a79f405ded94f875f868fe533e138ec5/src/agent-manager.ts#L212) uses to evict records past the [10-minute cutoff](https://github.com/tintinweb/pi-subagents/blob/0c88e522a79f405ded94f875f868fe533e138ec5/src/agent-manager.ts#L807) — the record is gone from the map, `session.dispose()` is never called, and nothing is emitted. The agent stays un-aborted and the ctx stays live, so the heartbeat this PR targets keeps running: one leaked session per evicted subagent.

The doc comment argues this matches [`AgentSessionRuntime.dispose()`](https://github.com/earendil-works/pi/blob/v0.80.6/packages/coding-agent/src/core/agent-session-runtime.ts#L390-L397), but there a hanging handler blocks a caller that is waiting on it; here it blocks nobody and strands the session instead. What I'd want: race the emit against a short timeout, then call `session.dispose()` either way.

**3. Smaller things.**

- Match pi's own guard in [`disposeSession`](https://github.com/tintinweb/pi-subagents/blob/0c88e522a79f405ded94f875f868fe533e138ec5/src/agent-manager.ts#L168-L183) so children with no handler stay fully synchronous — [`emitSessionShutdownEvent`](https://github.com/earendil-works/pi/blob/v0.80.6/packages/coding-agent/src/core/extensions/runner.ts#L189-L196) does this:
  ```ts
  if (runner?.hasHandlers("session_shutdown")) {
    await runner.emit({ type: "session_shutdown", reason: "quit" });
  }
  ```
- [`const runner = session.extensionRunner;`](https://github.com/tintinweb/pi-subagents/blob/0c88e522a79f405ded94f875f868fe533e138ec5/src/agent-manager.ts#L170) is outside the `try`. [Today's getter](https://github.com/earendil-works/pi/blob/v0.80.6/packages/coding-agent/src/core/agent-session.ts#L3243-L3245) is a plain field read so it can't throw; with both call sites using `void`, a future throwing getter becomes an unobserved rejection. Cheap to move inside.
- `runner.emit` already [catches handler errors](https://github.com/earendil-works/pi/blob/v0.80.6/packages/coding-agent/src/core/extensions/runner.ts#L757-L767) and routes them to `emitError`, so the `try/catch` around the emit only covers a throw from `emit` itself — the comment above it ("a throwing handler must not block disposal") describes something `emit` already guarantees.

**4. Layering — question, not a blocker.**

`AgentSessionRuntime` and `createAgentSessionRuntime` are exported from the package root, and in pi `session.dispose()` is only ever reached through them — both call sites emit first. Wrapping child sessions in the runtime would inherit that invariant permanently instead of maintaining a replica that has to track pi's changes (the `hasHandlers` guard, `beforeSessionInvalidate`, reason values). It isn't free: `createAgentSession` returns `{session, extensionsResult, modelFallbackMessage}` with no `services`, so it means going through `createAgentSessionServices`/`createAgentSessionFromServices`.

tintinweb:
fixed with https://github.com/tintinweb/pi-subagents/commit/9afe114c446c35804f3ca218bbc754cfb9620a95 ; credits in changelog, thx!.

PR #234, opened 2026-08-16
feat(ui): use primary text color for entire selected FleetView row

### LINKED ISSUE #230: feat(ui): use primary text color for entire selected FleetView row
## Summary

FleetView should render the entire selected row with the theme's primary `text` color. This makes the active row read as one selection instead of several unrelated emphasis levels.

## Current behavior

Only the filled selection bullet uses the accent color. Agent name remains muted, task text is bright, and elapsed/token stats remain dim.

Current FleetView selection

## Expected behavior

Use the theme's primary `text` color for every selected-row element:

- selection marker
- agent display name
- task description
- elapsed time and token count

Inactive rows should keep current muted/dim styling. This should use the theme token rather than hard-coded white, so light themes remain correct.

Reference behavior:

Reference selected-row treatment

## Implementation pointer

Relevant code: `src/ui/fleet-list.ts`

Potential approach:

- Change selected marker from `theme.fg("accent", "●")` to `theme.fg("text", "●")`.
- Branch in `renderAgentRow()` on `rosterIndex === sel`.
- Render selected agent name, description, and right-aligned stats using `theme.fg("text", ...)`.
- Consider temporarily suppressing configured agent badge colors while selected, keeping the whole row visually uniform.

## Environment

- `@tintinweb/pi-subagents`: `0.16.1`
- Pi: `0.84.2`
- Theme: dark `claude-code`

kunaaal13:
Fixes #230

The selected FleetView row now renders uniformly in the theme's primary `text` token — selection marker, agent display name, task description, and the right-aligned elapsed/token stats. Configured agent badge colors are suppressed while the row is selected so the row reads as one selection. Inactive rows keep the existing muted/dim styling. Uses the theme token (no hard-coded white), so light themes stay correct.

## Before / after

Tested live in pi 0.84.2, dark `claude-code` theme, 3 background agents running:

| Before | After |
|---|---|
| before | after |

## Changes

- `src/ui/fleet-list.ts`: selected marker `accent` → `text`; `renderAgentRow()` branches on selection and renders name, description, and stats with `theme.fg("text", ...)`, suppressing the agent badge while selected (plain `displayName` via `getConfig`)
- `test/fleet-list.test.ts`: one new test asserting the selected row uses the `text` token everywhere and unselected rows keep muted/dim

No `package*.json` changes. `tsc`, `biome check`, and the full vitest suite (1255 tests) pass.


src/ui/fleet-list.ts
```diff
@@ -368,14 +369,22 @@ export class FleetList {
   }
 
   private bullet(rosterIndex: number, sel: number, theme: Theme): string {
-    return rosterIndex === sel ? theme.fg("accent", "●") : theme.fg("dim", "○");
+    return rosterIndex === sel ? theme.fg("text", "●") : theme.fg("dim", "○");
```
tintinweb: should we keep the bullet itself "accent"?

kunaaal13: Yes we can.

PR #236, opened 2026-08-16
feat: Workflow tool - deterministic scripts that orchestrate many subagents

tintinweb:
## What this adds

A new `SubagentWorkflow` tool. You give it a small JavaScript script, and it spawns and coordinates many subagents for you: fan out over a list, run each item through several stages, verify each result, and return a summary.

Until now the model could only start a handful of agents by naming them one by one in a single message. That works for three agents you already know about. It does not work for "audit every route file in this repo", where the list only exists once something has gone and looked. A script can loop, branch, and fan out over a list discovered at runtime; a batch of tool calls cannot.

**On by default.** Offering the tool costs a tool spec, not a fan-out, so `workflowsEnabled` now defaults to on instead of requiring opt-in.

Unset means **auto**: on, unless another loaded extension already provides a `Workflow` or `SubagentWorkflow` tool. In that case this one warns — naming that extension and the setting that overrides it — and stands down for the session, so a session never runs two competing orchestrators. Setting `workflowsEnabled` explicitly pins it in either direction; only the default defers. A stand-down is session-scoped and is never persisted back as an explicit `false`.

When off, the tool is never registered: the model is not told it exists and it costs no tool-spec context, the `/agents → Workflows` entry is hidden, and `--subagents-workflow-file` refuses with a pointer to the setting. Toggle it in `/agents → Settings → Workflows` or in `subagents.json`; the value is read at extension load, so it applies from the next pi session.

The collision check runs at `session_start` through `setActiveTools` rather than gating registration: `getAllTools` throws during extension loading, and load order means a registration-time check could not see an extension that has not loaded yet. `setActiveTools` rebuilds the system prompt before any turn runs, so the model never sees a spec that was withdrawn. Matching is exact against those two tool names, never a substring.

> [!IMPORTANT]
> **Requires pi 0.84.0 or newer.** `peerDependencies` moves up from `>=0.81.0`, so npm flags an older pi at install time. `SubagentWorkflow` needs two host APIs that do not exist below that floor, and both are typecheck failures there rather than features that quietly do nothing: `constrainedSampling` on pi's `ToolDefinition` (0.82.0), and pi-tui's `stripTerminalSequences` (0.84.0), which the workflow card and the inspector use to strip the resets `truncateToWidth` wraps its ellipsis in. Reimplementing the latter would have held the floor at 0.82, but it has to match the host's own truncation exactly to stay correct, and duplicating a host utility to buy back two minor versions is the worse trade. A CI job installs 0.84.0 and runs the full typecheck and suite against it, since a green run there is the only evidence behind the range.

Also in this PR:
- `/agents → Workflows`, a live inspector with pause / resume / skip / retry / stop
- a `workflow` row in FleetView for each running run
- saved workflows (`name:`), script files (`scriptPath:`), and journal-backed `resumeFromRunId`
- `pi --subagents-workflow-file=<path>` to run a workflow at startup, including headless
- worktree isolation no longer freezes the UI while git runs
- `agent({ model })` takes the same `enabledModels` scope check the Agent tool does, and a row reports the model the child actually got rather than the one the script asked for
- `docs/workflows.md`, and a workflows section in the README
- the pi peer floor moves to `>=0.84.0` (see above)

## How you use it

You mostly don't write the script yourself. You ask for the work, and the model writes one. Here is what it writes:

```js
export const meta = {
  name: 'auth-audit',
  description: 'Find routes missing auth checks, then verify each finding',
  phases: [{ title: 'Scan' }, { title: 'Audit' }],
}

phase('Scan')
const listing = await agent('List every route file under src/routes/. One path per line.')
const files = listing.split('\n').map(s => s.trim()).filter(Boolean)
log(`auditing ${files.length} files`)

phase('Audit')
return await pipeline(
  files,
  file => agent(`Audit ${file} for missing auth checks.`, { label: `audit:${file}` }),
  (found, file) => agent(`Try to REFUTE this finding about ${file}: ${found}`, { label: `verify:${file}` }),
)
```

One agent finds the files. Then every file gets audited and independently re-checked, all at once, capped so your machine stays usable.

The tool returns a task id straight away and the run continues in the background, notifying through the same channel background agents already use. You get a card in the chat showing phases and agents as they go.

The `meta` block at the top must be a **pure literal** — no variables, function calls, spreads or template strings. That is what lets it be read *before* the script runs, so the phases are on screen from the first frame instead of appearing one at a time as agents happen to start.

### Tool parameters

| | |
|---|---|
| `script` | Inline source |
| `scriptPath` | A script file, absolute or project-relative. Wins over `script` — this is how you re-run an edited workflow |
| `name` | A saved workflow: `<name>.js` in `.pi/workflows/`, `.agents/workflows/`, or `<agent dir>/workflows/`. Lowest precedence |
| `args` | Handed to the script as the `args` global, verbatim |
| `resumeFromRunId` | Replay an earlier run in this session — see [Resume](#resume) |
| `title` / `description` | Accepted and ignored, as in Claude Code. A workflow is named by its `meta` block |

Every invocation persists its script to the session directory and returns the path, so iterating means editing that file and re-running rather than resending the source. A saved workflow reports its own file instead, so the same loop works on it. Those directories are ordinary folders that may hold other scripts, so only files carrying the `export const meta = { name, description }` declaration are listed or resolved — naming anything else says it is not a workflow rather than producing a parser error about a block the author never meant to write. The check is a regex; nothing in the file is executed to make it, and even the real parse evaluates only the `meta` object literal, in an empty vm with a 100ms bound.

### What the script can call

| | |
|---|---|
| `agent(prompt, opts?)` | Spawn one subagent, get its final text — or, with `schema`, a validated object. `null` if it failed terminally or you skipped it |
| `pipeline(items, ...stages)` | Push every item through every stage, with no waiting between stages |
| `parallel(thunks)` | Run things at once and wait for all of them |
| `workflow(nameOrRef, args?)` | Run a saved workflow inline as a sub-step |
| `phase(title)` | Start a new group in the progress display |
| `log(message)` | A progress line for you to read |
| `args` | Whatever was passed to the tool, verbatim |
| `budget` | `{ total, spent(), remaining() }` — see below |

`agent()` options: `label`, `phase`, `agentType`, `model`, `effort`, `isolation: "worktree"`, `gate`, `resume`, `schema`. Any other key is rejected by name at the call rather than silently ignored.

A `pipeline` stage receives `(previousResult, originalItem, index)`, and a stage that throws drops that one item to `null` and skips its remaining stages. In `parallel`, a thunk that throws becomes `null` without taking its siblings down.

### pipeline vs parallel

This is the one thing worth knowing. `parallel` waits for everything before anything moves on. If five agents run and the slowest takes three times as long as the fastest, four of them sit finished doing nothing.

`pipeline` doesn't wait. One file can be in the verify stage while another is still being audited, so total time is the slowest single chain rather than the sum of the slowest per stage. Use `parallel` only when a step genuinely needs all the previous results together, like de-duplicating findings before deciding what to verify.

The model-facing tool description spends real space on this, because reaching for a barrier when you wanted a pipeline is the easiest mistake to make and the most expensive one.

### Getting an object back instead of prose

```js
const findings = await agent('Review src/auth.ts', { schema: FINDINGS_SCHEMA })
```

`schema` takes a JSON Schema whose root is an object. The child is given a `StructuredOutput` tool built from it, and `agent()` resolves to the **validated object** rather than text, so the script never parses prose. A payload that does not match is answered with `isError` and the reason, so the child corrects itself inside the same run; one that answers in prose anyway gets one further prompt, and one that still doesn't comply fails — so the call returns `null` rather than something unparseable. Mutually exclusive with `resume`, whose session predates the tool.

One honest gap: Claude Code *forces* the tool call, and this cannot. `toolChoice` exists in pi-ai's provider layer but isn't plumbed through `AgentSession`, so the pressure here is `constrainedSampling`, the tool's own description, validation, and the retry prompt — four soft pressures where Claude Code has one hard one. Scripts see the same types either way; a stubborn child costs a `null` rather than a wrong shape.

Validation uses `typebox` v1, not `@sinclair/typebox` — both are installed. The latter dispatches on a `Kind` symbol a wire-format schema does not carry and throws `Unknown type` on plain JSON Schema; v1 is a standards validator and is what pi types `ToolDefinition.parameters` against, so one schema object serves as both the tool's parameters and the check.

### Checking work by running a command, not by asking a model

```js
agent('Fix the failing test in src/auth.ts', { gate: 'npm test' })
```

`gate` runs the command after the agent finishes and fails the agent on a non-zero exit, with the command's output as the error. Asking a second model "does this look right?" is a weak signal; running the test suite is not.

The failure reuses the shape an agent failure already had, so the card, the inspector and the `null` returned to the script all handle a gate rejection with the code they already had — no new state anywhere.

If the agent worked in an isolated worktree, **the gate runs inside that worktree**. That needed a new `onBeforeWorktreeCleanup` hook on the manager: the settle path commits the copy to a branch and removes it before the spawn resolves, so a gate that looked at the path afterwards would quietly check your main tree instead, and could pass while the agent's actual changes were broken.

### Picking up where an agent left off

```js
agent('Here is what the tests said, fix it', { resume: 'fix-auth' })
```

`resume` continues the agent that ran under that label instead of starting a fresh one that has to re-read everything. A resumed agent keeps its stored agent type, model and tree, so passing `agentType`, `model` or `isolation` alongside `resume` is rejected rather than silently ignored.

An agent whose gate failed stays resumable, which is what turns gate + resume into a real retry loop.

### Composing workflows

```js
const map = await workflow('repo-map', { root: 'src' })
```

`workflow(nameOrRef, args?)` runs a saved workflow inline and returns its value; pass a name or `{ scriptPath }`. It is not a second run — the child executes in the same worker and the same vm context under its own globals, so it shares this run's concurrency cap, agent counter, abort signal, resume journal and budget by construction rather than by plumbing. Its agents are simply this run's agents, numbered in the same sequence and controllable from the same dialog. What it does not share is phase state: a child's phases render as their own `▸ <name>` group, so its `phase("Scan")` never aliases the parent's.

Nesting stops at one level, and `workflow()` inside a child throws saying so. An unknown name, an unreadable path, a child with no `meta` or one that won't parse all throw into the calling script, which can `catch` them. Capped at 256 nested calls per run.

Reach for it to reuse a saved workflow, not to structure one script — inline composition is cheaper.

### `budget`

`{ total, spent(), remaining() }`. `total` is always `null` here: it comes from a token-target directive pi does not have, so Claude Code's `while (budget.total && budget.remaining() > 50_000)` guards correctly do not fire and such a script still runs. `spent()` is real — output tokens this run's agents have used — and `remaining()` is `Infinity` with no target.

### Forgotten `await`

If a script finishes while an `agent()` call is still running because someone dropped an `await`, the run fails instead of quietly throwing that agent's work away.

The check lives on the host side, since the runtime already tracks every launch, so it needs no `Promise` wrapper inside the sandbox. It fails immediately rather than waiting the stray agent out: an agent that ignores its abort signal would otherwise wedge the run's completion forever. A run you aborted on purpose doesn't report stragglers.

## Resume

Every run journals each settled `agent()` call next to its persisted script, as `<run id>.workflow.jsonl`. `resumeFromRunId` replays the **unchanged prefix** of that journal: the first call whose prompt or spawn options changed, and everything after it, runs live. Same script and args means nothing re-runs at all. Passing only `resumeFromRunId`, with no script of its own, re-runs that run's own script.

It is a prefix and not a lookup table on purpose. A later call that still matches came from a run whose earlier stages no longer exist — the same prompt at position 12 of a *different* run is not the same work, because what fed it changed.

A journaled failure is never replayed as one: resuming a run that died at agent 5 exists to retry agent 5, so the prefix ends there. A run that used `agent({ resume })` is not replayed at all — a replayed agent is text from a file, not a live child, so there is no conversation in the new run for a later `resume` to continue. Coarse on purpose; the alternative is a second key concept for a case that costs one run.

Same session only, and the run must have finished (stop it from `/agents → Workflows` first). Replayed rows are annotated `from resume journal` in both the card and the dialog, and the completion notification says how many there were.

## `/agents → Workflows`

The inspector lives in the agents menu rather than as a top-level `/workflows` command: it is one more view of the same fleet, and a second command name only adds a collision surface — pi renames duplicates to `/workflows:1` and `/workflows:2`, which breaks the bare name for both. A run is also reachable by selecting its row in FleetView.

```
 audit-src
 Dynamically discover files under src/ and audit each …                    1/3 agents · 32s

 ╭ Phases ──────────┬ Discover · 1 agent ──────────────────────────────────────────────╮
 │ ❯ ✔ Discover 1/1 │ ❯ ✔ discover:src Opus 5 (1M context) · 26.4k tok             25s │
 │   2 Audit    0/2 │                                                                  │
 │   3 Verify       │                                                                  │
 │   4 Synthesize   │                                                                  │
 ╰──────────────────┴──────────────────────────────────────────────────────────────────╯
 ↑↓ select · ⏎ open · f filter · x stop · esc close
```

Two panes, two levels. The overview puts the phases on the left — a phase shows its number until it finishes, then `✔` or `✘` — and the selected phase's agents on the right. `⏎` opens one: the agents move left and the right pane becomes that agent's **Prompt**, **Activity** and **Outcome**, with `⏎` now expanding the prompt and `esc` going back a level rather than closing.

| Key | |
|---|---|
| `↑↓` / `j` `k` | Move |
| `⏎` | Open the phase, then expand the prompt (`e` also toggles it) |
| `f` | Cycle the state filter, named in the pane title |
| `x` | Stop the run — live runs only |
| `p` | Pause / resume. Pausing stops *starting* agents and leaves running ones to finish, because killing model work mid-turn throws away everything it has spent. Held time is subtracted from the elapsed clock |
| `s` | Skip the selected agent: its `agent()` call returns `null`, exactly as a terminal failure does |
| `r` | Retry the selected agent: the child is stopped and the same call runs again, so the script's promise is still the one waiting. Running agents only — a settled call's value is already the script's, and a re-run would have nowhere to put one. The row then reads `attempt 2 · user retry` |
| `esc` | Back a level, then close |
| `ctrl+c` | Close outright, from either level. Deliberately not folded into the `esc` branch — stepping back a level on the reflex key leaves the overlay on screen |

The footer offers only keys that are wired *and* currently actionable, so it never advertises an action that does nothing. Skipping is immediate for a running agent and for one held at a pause; one parked behind the concurrency limit takes its skip when it reaches the front of the queue. If more than one workflow has run this session, it asks which one, newest first.

## FleetView

A running workflow is one `workflow` row above the agents, carrying its agent counts where a description would go; `Enter` opens the inspector rather than a conversation overlay.

A run's own agents are **not** listed separately — they belong to the run, which reports for them, so they are filtered out of the fleet list, the above-editor widget, the `/agents` menus and `@handle` resolution exactly as nested children are. They are also outside the `maxConcurrent` pool — and outside `maxConcurrentForeground`, the second pool master grew in 0.18.2 — because the run has its own cap, and charging its agents to a session limit as well would let one fan-out queue behind, or starve, everything else you are doing.

## Running one from the command line

```
pi --subagents-workflow-file=review.js
pi -p --subagents-workflow-file=review.js      # headless
```

Use the `=` form. The bare `--flag value` spelling swallows the next argument, so a flag written before a prompt would eat the prompt.

The flag is registered when the extension activates but read at `session_start`, because pi only applies CLI values after every extension has loaded — read any earlier and you get the registered default instead of what the user typed. With workflows off it refuses and names the setting, rather than doing nothing.

With no tool call to attach to, the run renders as a normal session entry using the same layout as the tool result, and its outcome is handed to the model as context for your next message rather than forcing a turn.

## Things scripts deliberately cannot do

Scripts run in a `node:vm` context on a worker thread. Inside a script:

- `Date.now()`, `new Date()` and `Math.random()` throw. A script that gives a different answer every run can't be replayed from its journal. Pass a timestamp in through `args`, or stamp it after the workflow returns
- `eval` and `Function(...)` throw `EvalError`, because code generation is switched off in the vm
- no filesystem, network or module access

That last one matters less than it sounds: the script is only the coordinator, and all the real work happens in the agents it spawns, which have their normal tools.

Two details behind the sandbox. `Object`, `Array`, `JSON` and `Promise` come from the vm's own realm rather than the host's, because a host built-in's `.constructor` is the host `Function` and hands the script a way back out. And values crossing in or out are screened for cycles, non-finite numbers, sparse arrays, symbol keys and odd prototypes, so nothing malformed reaches the progress log or the model.

To be clear about what all of this is for: it stops accidents and non-determinism. It is not a defence against a deliberately hostile script — the injected globals are host closures, and disabling code generation is what actually prevents one being used to compile anything.

Limits per run: concurrency at `max(1, min(16, cpus - 2))` — the `max(1, …)` is load-bearing, since plain `min(16, cpus - 2)` is `0` on a two-core machine and a semaphore with zero permits never hands out a slot — plus 1000 agents per run, 4096 items per `parallel`/`pipeline` call, and 256 nested `workflow()` calls.

## Compatibility with Claude Code

This is a port of Claude Code's `Workflow` tool down to its internal state model, so **a script written for Claude Code runs here unchanged**. `test/workflow-claude-code-compat.test.ts` runs the canonical `review-changes` example from that tool's own description, verbatim — `schema`, `pipeline`, `parallel`, `phase`, template-literal labels and `.then` chaining inside a stage, all at once.

The model-facing description is that tool's own wording, carried over verbatim and deviating only where a sentence would be false here: the tool's name, the settings and menus it points at, `budget.total`, and where the resume journal lives. A run's children are told through a `<workflow_child>` block in their system prompt that their final text *is* the return value rather than a message for a person to read, which is what keeps a stage without `schema` returning data instead of prose.

Progress is an append-only event log rather than a tree. Agent entries are keyed by index and the last write wins, so a running agent is updated by appending a fresh entry with the same index instead of mutating anything, and every view — the inline card, the dialog, the fleet widget — gets its shape by collapsing that log. That's what lets the worker batch several agents' changes into one message.

`gate`, `resume`, `effort`, the forgotten-`await` check and journal-backed resume are the additions. All are optional, which is what keeps a Claude Code script portable, and `test/workflow-borrowed.test.ts` exists to keep it that way as much as to test the features. The differences that remain are named above: `schema` is pressured rather than forced, and `budget.total` is always `null` because pi has no token-target directive.

## Worktree isolation no longer blocks the UI

Every git call in `worktree.ts` moved from `execFileSync` to `pi.exec`. They ran on the same thread as the terminal UI with timeouts up to ten seconds, so `git worktree add` on a large repo froze the whole interface. Workflows make that worse, since a fan-out of isolated agents turns one blocking call into several back to back.

From the outside nothing changes. `AgentManager.spawn()` keeps its synchronous signature and still returns an id immediately, which the tool, the scheduler and the cross-extension RPC all rely on. Worktree creation moved into the async run path instead, behind a new `awaitStartup(id)` that resolves once the run is underway and rejects with the same strict-isolation error `spawn()` used to throw.

Three things kept working through the move:

- failures behave as before on both paths — an immediate spawn drops the record so no orphan shows up in `listAgents()`, while a queue-drained one records the error
- the pool slot is claimed *before* the awaited copy, not after. Otherwise `drainQueue`'s synchronous loop would start every queued agent at once, and `abort()` couldn't reach one mid-start
- `pi.exec` reports a timeout as exit code 0 with `killed` set, so that flag is checked separately. It's the one failure that would otherwise read as success

## Layout

```
src/workflow/
  meta.ts             # Extract and validate a script's pure-literal `meta` block
  worker-source.ts    # The sandbox: vm context, determinism prelude, script globals
  runtime.ts          # Worker lifecycle, RPC bridge, semaphore, caps, gate/resume/schema
  progress.ts         # Progress event log and every derived view of it (pure)
  host.ts             # WorkflowHost adapter over AgentManager
  task.ts             # local_workflow task record and batched progress updates
  journal.ts          # Per-run JSONL record behind resumeFromRunId
  saved.ts            # Resolve `name:` to a script under .pi / .agents / agent dir
  json-schema.ts      # Validate a script-supplied JSON Schema (typebox v1)
  tool-description.ts # Model-facing description carrying the orchestration patterns
src/
  structured-output.ts  # The synthetic StructuredOutput tool behind `schema`
src/ui/
  workflow-card.ts    # Inline workflow card (tool result and session entry)
  workflow-dialog.ts  # The two-pane inspector
```

Spawning is injected into the runtime rather than imported. `AgentManager` is large and stateful, and wiring it in directly would make every runtime test an integration test; a `WorkflowHost` stub is a dozen lines. `progress.ts` is pure and framework-free, so the whole display model is unit-testable without a terminal.

## Tests

About 7.4k lines of source and 9.3k of tests across the branch: `meta` extraction, the runtime, the sandbox, the progress model, rendering, the inspector, the tool itself, structured output, JSON Schema validation, the resume journal, saved workflows, gate-inside-worktree, the CLI flag, the Claude Code compatibility script, and an end-to-end run. The existing worktree and agent-manager tests were reworked for the async spawn path.

`npx vitest run` — 104 files, 2112 passed, 4 skipped. Rebased onto `master` at v0.18.2.


tintinweb:
name: /SubagentWorkflow

(not folded into /Agent bcause it may confuse the agent; not /Workflow because collision + collision detection)

tintinweb:
haiku likes it, but haiku likes everything :D 

image
image
image
image

PR #264, opened 2026-08-26
fix: bound conversation viewer render cost

### LINKED ISSUE #263: Conversation viewer cap still scans full output and retries failed Markdown
Follow-up to #210 / #259. Verified against current `master` at `bd446fcce2b7b36778127a7877fd13df7088e9e5` (v0.19.0).

The conversation viewer has two paths where its 16,000-code-unit result cap and Markdown failure cache do not bound render work as intended.

## 1. Truncation scans and allocates across the full hidden tail

`src/ui/conversation-viewer.ts` currently computes:

```ts
elided: text.slice(RESULT_MAX_CHARS).split("\n").length
```

`buildContentLines()` runs from both `render()` and scroll-bound calculation in `handleInput()`. The original string is already in memory and V8 may represent `slice()` without copying, but `split()` still scans the entire hidden tail and allocates its segment array. The visible result is capped while render/scroll work remains O(total result size).

### Reproduction evidence

Environment: M1 Max, Node 22.23.1, Vitest 4.1.10. One default-mode ASCII `toolResult`, approximately one newline per 22 code units, width 120. The current implementation and an O(1) omitted-count variant were alternated for 7 rounds with GC before each round.

Median time per operation:

| Total result size | Current render | O(1) render | Current scroll | O(1) scroll |
| --- | ---: | ---: | ---: | ---: |
| 16,000 | 0.408 ms | 0.411 ms | 0.231 ms | 0.235 ms |
| 1 MiB | 1.511 ms | 0.437 ms | 1.180 ms | 0.258 ms |
| 8 MiB | 14.018 ms | 0.554 ms | 13.805 ms | 0.311 ms |
| 32 MiB | 70.054 ms | 0.735 ms | 74.661 ms | 0.451 ms |

All 42 comparisons with a hidden tail were slower in the current implementation. At 32 MiB, observed net heap growth was about 58 MiB for current render/scroll versus under 0.25 MiB for the O(1) variant; GC makes this a lower-bound observation rather than an exact allocation total.

A bounded alternative is to report omitted UTF-16 code units in O(1):

```ts
elided: text.length - RESULT_MAX_CHARS
```

and render `... (truncated, N more characters)`. Exact omitted-line counts would need to be maintained outside the render path. The current `split()` result is a tail-segment count and is not exact at every cut boundary.

## 2. Append-only streaming retries a known failing Markdown prefix

After `Markdown.render()` throws, `entry.failed` correctly selects literal wrapping. However, every text update resets it:

```ts
entry.md.setText(text)
entry.text = text
entry.failed = false
```

The documented failure case is a deeply nested blockquote prefix that overflows the recursive renderer. Appending a streaming delta does not remove that prefix.

A viewer test harness forced `Markdown.render()` to throw, then appended five deltas to the same message. Across 50 independent viewers, the current implementation attempted and caught Markdown failures with cumulative counts `1,2,3,4,5,6`; the append-aware variant stayed at `1,1,1,1,1,1`. Every appended delta remained visible through literal fallback in both cases.

Keep the failure cached for append-only growth and retry when content was replaced or shortened:

```ts
const shouldRetry = !text.startsWith(entry.text)
entry.md.setText(text)
entry.text = text
if (shouldRetry) entry.failed = false
```

This deliberately does not retry a failure that could hypothetically become valid through appending. A focused regression should count Markdown render attempts: force one failure, append to the same message five times, render after each append, and assert the attempt count stays at one while all appended text remains visible through literal fallback.

No open issue or PR among the current upstream set fixes either path.

HerbertGao:
Fixes #263

## Summary

- make the 16,000-code-unit viewer cap report omitted UTF-16 code units in O(1), instead of splitting and allocating across the entire hidden tail during every render and scroll-bound calculation
- keep Markdown's literal fallback cached across append-only streaming deltas after a render failure
- retry Markdown when the message content is replaced or shortened, so a corrected message does not stay literal forever
- use the concrete `AgentSession` message type for the Markdown cache key

## Evidence

The issue benchmark reproduced the hidden-tail cost in all 42 comparisons. At 32 MiB, median render time fell from 70.054 ms to 0.735 ms and scroll handling from 74.661 ms to 0.451 ms with the O(1) count.

The failure harness reproduced 50/50 times: one initial failure plus five append-only deltas caused six caught Markdown throws before this fix and one after it.

## Verification

- focused conversation-viewer suite: 59 passed
- `npm run lint` (only the existing `workflow-tool-description.test.ts` template-placeholder warning on master)
- `npm run typecheck`
- `npm run test`: 105 files, 2126 passed, 7 skipped
- `npm run build`
- active LSP/Lens diagnostics: no errors in the two changed files

Per CONTRIBUTING.md, this PR does not edit CHANGELOG.md.

### DIFF
--- a/src/ui/conversation-viewer.ts
+++ b/src/ui/conversation-viewer.ts
@@ -28,8 +28,8 @@ export const VIEWPORT_HEIGHT_PCT = 70;
  * every render *and* on every scroll key (`handleInput` calls it to compute
  * `maxScroll`), so an uncapped 200 KB result costs ~6 ms per keystroke to parse
  * as Markdown, against ~0.5 ms once capped and effectively nothing on a cache
- * hit (best of 5, width 76). 16,000 code units is roughly a screenful at every
- * terminal size and still ~30x the 500 characters this replaces, which was small enough to cut
+ * hit (best of 5, width 76). 16 KB is roughly a screenful at every terminal size
+ * and still ~30x the 500 characters this replaces, which was small enough to cut
  * most real results mid-sentence.
  */
 export const RESULT_MAX_CHARS = 16_000;
@@ -121,7 +121,7 @@ function capResult(text: string): { text: string; elided: number } {
 }
 
 function truncationNote(elided: number): string {
-  return `... (truncated, ${elided} more UTF-16 code unit${elided === 1 ? "" : "s"})`;
+  return `... (truncated, ${elided} more character${elided === 1 ? "" : "s"})`;
 }
 
 export class ConversationViewer implements Component {
--- a/src/ui/conversation-viewer.ts
+++ b/src/ui/conversation-viewer.ts
@@ -120,8 +120,21 @@ function capResult(text: string): { text: string; elided: number } {
   };
 }
 
+/**
+ * `999` · `1.5k` · `8.4M` — a magnitude cue, not an exact count, past 1000.
+ *
+ * The bracket is chosen against the *rounded* value, so 999,999 reads `1M`
+ * rather than the `1000.0k` a naive `< 1e6` test produces.
+ */
+function humanCount(n: number): string {
+  if (n < 1_000) return `${n}`;
+  const thousands = n < 999_950;
+  const value = thousands ? n / 1_000 : n / 1_000_000;
+  return `${value.toFixed(1).replace(/\.0$/, "")}${thousands ? "k" : "M"}`;
+}
+
 function truncationNote(elided: number): string {
-  return `... (truncated, ${elided} more character${elided === 1 ? "" : "s"})`;
+  return `... (truncated, ${humanCount(elided)} more character${elided === 1 ? "" : "s"})`;
 }
 
 export class ConversationViewer implements Component {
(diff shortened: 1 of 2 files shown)


tintinweb:
Good catch. Before = bd446fc, after = this branch.

Elision count. buildContentLines() runs on every render and every scroll keystroke, so the split() was costing input lag, not just background CPU:
```
┌────────────────┬──────────┬─────────┐
│  result size   │  before  │  after  │
├────────────────┼──────────┼─────────┤
│ 16 KB (at cap) │ 0.54 ms  │ 0.54 ms │
│ 1 MiB          │ 8.47 ms  │ 0.54 ms │
│ 4 MiB          │ 38.04 ms │ 0.52 ms │
│ 8 MiB          │ 98.39 ms │ 0.60 ms │
└────────────────┴──────────┴─────────┘
```

one change: character count with short version amount (1000 -> 1k ...)
```
notice                                  cols   at width 50 (46 usable)
1100000 more UTF-16 code units           47    ...more UTF-16 code un...  ← unit cut off
1.1M more characters                     37    fits
```

PR #265, opened 2026-08-27
fix: hold completion notifications until the parent run settles

vincelwt:
## Summary

Background completion notifications are sent with `deliverAs: "followUp"`. Pi only delivers those once the parent has no more tool calls, so a notification emitted mid-run parks in pi's follow-up queue and cannot be withdrawn.

That means an agent already joined with `get_subagent_result` (or `subagents:rpc:consume`) still notified after the final answer. With `followUpMode: "one-at-a-time"` those stale messages drain one wasted turn each. Observed in the wild: seven agents that had finished up to 1h41m earlier flushed in 54 seconds after the final answer and forced a compaction.

This keeps due notifications in-process instead:

- 200ms hold so a same-tick join can still cancel
- if the parent is mid-run, park until `agent_settled` (not `agent_end` / `turn_end`, which still have work ahead)
- send closures already re-check `resultConsumed`, so deferring the call defers the check
- busy state is read from `ctx.isIdle()` at delivery time

Idle parents still notify immediately. Shutdown drops undelivered notifications, matching `abortAll()`.

## Verification

- `npm run lint`
- `npm run typecheck`
- `npm run test` (2140 passed, 7 skipped)
- `npm run build`

### DIFF
--- a/README.md
+++ b/README.md
@@ -582,6 +582,8 @@ When background agents complete, they notify the main agent. The **join mode** c
 
 **Timeout behavior:** When agents are grouped, a 30-second timeout starts after the first agent completes. If not all agents finish in time, a partial notification is sent with completed results and remaining agents continue with a shorter 15-second re-batch window for stragglers.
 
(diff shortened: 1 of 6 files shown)


vincelwt:
Closing. Not needed on 0.19.0.

The hold only matters if a result is consumed more than 200ms after completion while the parent is still running. Current Agent guidance is not to join: spawn, end the turn, wait for the notification. Same-tick cancel plus `subagents:rpc:consume` already cover the remaining consume paths.

PR #289, opened 2026-09-04
feat(ui): add session/prompt/info/output agent views

### LINKED ISSUE #288: Agent viewer: Session / Prompt / Info / Output panes for faster readability
# Problem

The only way to inspect a subagent is the conversation overlay. It opens on the live transcript and auto-scrolls to the end. For me, navigation in this transcript view is usually somewhat laggy.

- **The spawn prompt.** It is the first user message, so it sits at the top. With `inherit_context`, it is also mixed into the parent conversation. There is no way to see “what did I hand this agent?” on its own.
- **The final output.** It is buried in tool calls and results. `get_subagent_result` exists for the model, not for me while I am looking at the overlay.
- **Run facts.** Model, thinking level, cost, tool count, elapsed time, and wall-clock start/end share one truncated header. Turns in the widget are a different surface again.

`/agents` → Running agents and FleetView Enter both jump straight into that overlay. Same problem from both entry points.

I want it to be very easy to see the subagent's input prompt (so I can diagnose how well my parent agent is prompting its subagents) and to see the subagent's final output.

# Proposed solution

One overlay, four views, switched in place:

## `/agents` → Running agents

1. Pick an agent.
2. A picker lists Session, Prompt, Info, Output. Session is the existing view.
3. Arrows + Enter, or press `s` / `p` / `i` / `o`.
4. Overlay opens on that view. Same keys switch views without closing it. Esc closes.

## FleetView

- Enter on an agent row opens the same picker (arrows + Enter, or the keys).
- `s` / `p` / `i` / `o` on a selected agent row skip the picker and open that view.
- Those keys do nothing on main or a workflow row.
- Inside the overlay, the keys work the same as from `/agents`.

Store the spawn prompt on the record at spawn time so Prompt does not have to reconstruct it from session messages.

I have this working locally and can follow with a PR.

HugoReeves:
Closes #288. States as of opening.

## Summary

The conversation overlay is the only way to inspect a subagent. It always opens on the live transcript and auto-scrolls to the end.

That hides three things:

- The spawn prompt sits at the top. With `inherit_context`, it is mixed into parent conversation.
- The final result is buried in tool calls and tool results.
- Model, thinking, cost, tool count, and times share one truncated header.

`/agents → Running agents` and FleetView Enter both jump straight into that overlay.

## What changed

One overlay, four views, switched in place with `t` / `i` / `p` / `o`:

- **Transcript** (`t`) — current live conversation (steer, stop, `m` unchanged)
- **Info** (`i`) — model, thinking, cost, tools, elapsed duration, start/end. No turn count
- **Prompt** (`p`) — the spawn `prompt` argument only, not parent context
- **Output** (`o`) — `record.result`, or `No output yet.` while the agent is still running

`t` is Transcript so it does not collide with workflow Skip (`s`). Overlay `p` is Prompt; the workflow inspector's `p` remains pause — they never share a key stream.

`/agents → Running agents` and FleetView Enter open a picker first (arrows + Enter, or the same keys). On a selected FleetView agent row, `t` / `i` / `p` / `o` skip the picker and open that view.

The spawn prompt is stored on `AgentRecord` at spawn time so Prompt does not reconstruct it from session messages.

## Related work

| Number | Title | State | Relation |
|---|---|---|---|
| #288 | Agent viewer: Session / Prompt / Info / Output panes for faster readability | open | This PR |

## Behavior and compatibility

No settings. No turn tracking. `main` and workflow rows in FleetView are unchanged. The workflow inspector `c` key still opens Transcript; the new keys work once that overlay is open.

Prompt is empty (`No prompt stored.`) for records spawned before this change, in the same session only until they are evicted.

Not a breaking change for the Agent tool or for session files.

## Performance

Not measured. `npm run bench` was not run. Prompt and Output reuse the existing `RESULT_MAX_CHARS` cap from Transcript tool results.

## Testing

```text
npm run lint        # biome check src/ test/ — no findings
npm run typecheck   # tsc --noEmit — pass
npm run test        # 106 files, 2147 passed, 7 skipped
```

New coverage: view picker keys (`t/i/p/o`, order Transcript/Info/Prompt/Output), spawn-prompt storage, Prompt vs parent context, Info without a turn count, Output empty state, overlay view switching including composer swallow and stop-disarm, FleetView Enter → picker, FleetView `p` on an agent row, FleetView `p` on `main`.

Not covered: live TUI `/agents` menu (headless tests cannot drive overlay stacking end-to-end). Mutation-testing of the new assertions was not run.


HugoReeves:
I will remove the docs file before merge.

HugoReeves:
Changed from Session (S keybind) to Transcript (t) to help avoid any annoying keymap confusion with the stop keybind in workflow view.

Issue #273, opened 2026-08-29
Group completion notifications ignore resultConsumed: 200ms suppression window, group nudge cannot be cancelled


Hayston1001:
## Summary

Group delivery path: a background agent whose result was already retrieved via `get_subagent_result` still produces a completion notification. The group nudge can **never** be proactively cancelled, and its only suppression window is 200 ms — so any consumption that happens after enqueue (e.g. an agent completing between two parent tool calls) always leaks a redundant follow-up turn.

Related but distinct from #185 / #186: those cover the *individual* notification arriving after the final answer; this one is the *grouped* notification branch, whose suppression is strictly weaker and unconditionally racy.

## Repro (real session)

- Two background agents spawned in one assistant message (smart join mode → registered as one group, `src/group-join.ts`).
- Both completed within ~300 ms of each other.
- Parent fetched the first result with `get_subagent_result(wait: true)` → resumed at completion, marked `resultConsumed = true`, its notification suppressed.
- Parent fetched the second result a few seconds later (model tool-call latency, ≫ `NUDGE_HOLD_MS`).
- **Result:** the group notification still arrived as a follow-up turn after the parent's turn had ended, with header:

```
Background agent group completed: 1 agent(s) finished
```

The "1 agent(s)" count is the smoking gun: the send-time re-check (`src/index.ts:505-509`) saw exactly one unconsumed record when the nudge fired, even though **both** records were consumed by the time the message was delivered. It then cost a wasted follow-up turn ("already incorporated; no further action needed").

## Root cause

1. **The group nudge can never be cancelled.** `get_subagent_result` calls `cancelNudge(params.agent_id)` (`src/index.ts:2804`), but the group nudge is keyed `group:${ids}` (`src/index.ts:501`). The keys never match, and there is no `agentId → groupKey` lookup anywhere. Group suppression therefore depends **entirely** on the send-time re-check.
2. **The re-check window is 200 ms** (`NUDGE_HOLD_MS`, `src/index.ts:451`), while the failure mode is model tool-call latency — typically 1–3 s. The window only covers "agent completed while the parent was already blocked in `get_subagent_result(wait: true)`", not the more common "agent completed between two tool calls".
3. Once the re-check passes, the notification is enqueued into pi's `followUp` queue and can no longer be withdrawn (`deliverAs: "followUp", triggerTurn: true`), so it drains after the run ends — same multi-minute delay as #185, one enqueue earlier.

[...]

PR #303, opened 2026-09-09
fix: hold grouped completion notifications until the parent run settles

Hayston1001:
## Summary

Follow-up to #273 (analysis + design discussion in the thread). Grouped
completion notifications were handed to pi's followUp queue 200 ms after the
last group member completed; any `get_subagent_result` after that point still
produced a stale notification turn ("Background agent group completed:
1 agent(s) finished" for a fully-joined group). The group nudge could never be
cancelled proactively: `get_subagent_result` cancels by agent id while the
group nudge is keyed `group:<ids>`.

pi exposes no per-message withdrawal once a custom message is enqueued
(`sendCustomMessage` hands it straight to the agent's queue; the only related
API is the all-or-nothing `clearQueue()`), so the delivery decision is moved to
the last responsible moment instead of widening the fire-time window:

- **Parent idle** → unchanged: 200 ms hold, per-record `resultConsumed`
  re-check at fire time, immediate delivery.
- **Parent mid-run** → the batch is parked in-process instead of calling
  `pi.sendMessage`.
- **Flush at `agent_settled`** (fires only after retries, auto-compaction, and
  queued follow-ups are done): re-check `resultConsumed` **per record** —
  fully-consumed batches are dropped, partially-consumed ones deliver only
  their remaining members with a recomputed count.

## Properties

- **No timeliness regression vs master**: a followUp parked in pi's queue
  today also lands at run end; flush-at-settled arrives at essentially the
  same moment. This removes the waste (stale content, wasted turns), not the
  timing.
- **Policy alternative to #190, not a collision**: #190 trades retractability
  for mid-run promptness via steering; this keeps arrival where master has it.
  If steering lands instead, the park/flush layer is confined to one dispatch
  function and can be swapped without touching the per-record filter.
- **Shutdown**: parked batches are dropped at `session_shutdown` rather than
  delivered — a turn cannot usefully be started during teardown, and after a
  session replacement they would be stale. (One deliberate refinement from the
  issue comment, which said "flush on shutdown".)

## Scope

Grouped delivery only (the #273 group branch). The individual/async path is
untouched — that is #190's territory.

## Tests

`test/group-settle-flush.test.ts` (wiring tests through the real extension,
same harness as `wait-queued.test.ts`):

1. mid-run park + sequential full consumption → no notification (the #273 repro)
2. partial consumption → remaining members only, count recomputed, consumed
   member's data absent
3. park + no consumption → delivered at settle, `deliverAs`/`triggerTurn`
   contract unchanged
4. idle parent → immediate delivery (behavior unchanged)
5. idle path + consumption inside the 200 ms hold → still suppressed by the
   fire-time re-check
6. `session_shutdown` → parked batches dropped, never delivered stale

Addresses #273

---

CI note: `npm test` on Windows shows 37 failures across 12 files that
reproduce identically on pristine `upstream/master` (worktree gitdir links,
memory paths, print-mode e2e) — pre-existing environment issues, not touched
by this PR.


joshua-zyy (commented):
I found one lifecycle issue that needs to be resolved before this leaves draft.

The issue reports Pi 0.84.x. I also checked the relevant behavior in Pi 0.85.0 and the current 0.85.1 release; the ordering is unchanged.

Pi tears down a session in this order:

```text
agent.abort()
  -> agent_settled
  -> session_shutdown
  -> dispose
```

`agent_settled` is emitted after Pi sets its internal streaming state to false.

This means the current implementation can flush parked notifications from its `agent_settled` handler before the `session_shutdown` handler gets a chance to clear them:

```ts
pi.on("agent_settled", () => {
  parentRunning = false;
  flushParkedGroupNotifications();
});

pi.on("session_shutdown", () => {
  parkedGroupNotifications.length = 0;
});
```

The flush calls `pi.sendMessage(..., { deliverAs: "followUp", triggerTurn: true })`. When Pi is no longer streaming, that API takes the non-streaming `triggerTurn` path and can start a new agent turn during teardown.

The current shutdown test invokes `session_shutdown` before `agent_settled`, which is the reverse of Pi's actual teardown order. I reproduced the real order in a temporary test while keeping the existing "sendMessage must not be called" assertion; it fails because `sendMessage` is called once.

Please update the test to use the actual event order and add a teardown-safe boundary so a parked notification cannot trigger a turn during shutdown. The exact implementation could be a pre-shutdown marker or a deferred/cancellable settled flush, depending on the intended Pi integration.

The normal settle-time behavior otherwise looks correct.

Hayston1001:
Nice catch — confirmed and fixed in de25a07.

I verified the ordering against pi's source before changing anything:
`teardownCurrent()` awaits `session.abort()` before emitting
`session_shutdown`, and `_emitAgentSettled()` clears the run-active state
before extension handlers run — so the flush indeed took the
non-streaming `triggerTurn` path mid-teardown. My shutdown test invoked
the events in the reverse order — harness bug on my side: the fake `pi`
fires events in whatever order I call them, real pi does not.

The fix is two layers:

- `session_before_switch` (which fires before `teardownCurrent` on every
  replacement path) sets a teardown marker, and the settled flush skips
  delivery while it is set — deterministic, no race. One documented edge:
  a vetoed switch leaves the marker until the next `agent_start`, and
  batches parked during that in-flight run are dropped (defensible — the
  user was mid-navigation away).
- The flush is also deferred by one macrotask and cancelled in
  `session_shutdown` — the safety net for any path without an early hook.

Tests now drive the real event order (`session_before_switch` → the
abort's `agent_settled` → `session_shutdown`) and assert no delivery,
with explicit coverage for the normal deferred delivery and the vetoed
switch. Thanks again — exactly the kind of issue the draft PR existed to
surface.

joshua-zyy:
Thanks for the quick fix and for confirming the lifecycle details.

I rechecked `de25a07f` against the current Pi 0.85.1 lifecycle, including the real `agent_settled -> session_shutdown` ordering. The updated group tests pass locally (7/7 with the host-only `general-purpose` test agent name replaced by an available agent), and `tsc --noEmit` plus Biome pass. I also verified the quit-style path without `session_before_switch`: the deferred flush is cancelled by `session_shutdown` and sends nothing.

The teardown blocker is resolved from my side. The documented vetoed-switch behavior is clear as an intentional tradeoff, so I have no further blocker at this point.

PR #329, opened 2026-09-16
feat(viewer): render tool calls as folded blocks in one full-screen frame

gjuoun:


gjuoun:
Opened in error — this belongs in the fork (gjuoun/pi-subagents), not upstream. Closing.

PR #334, opened 2026-09-19
refactor(error-handling): adopt neverthrow Result in the model path and lib adapters

gjuoun:
No issue: the fork has issues disabled, so this is the queued follow-up to the src/ domain-layout reshape (#8), taken from its handover. It closes nothing.

## Summary

The reshape's handover left one objective queued: the code reads as verbose and machine-written, and error handling was the largest single source of that — 41 throw sites, 104 try blocks of which 66 swallowed into a fallback value, and three spellings of the same hand-rolled ok-union (ok, success, kind). Three of those habits were shipping as real defects:

- **A corrupt pi settings.json silently switched scopeModels enforcement off.** The reader could not tell a file it failed to parse from one that was never configured, so both read as "no allowlist": the guard disabled itself while the setting still showed "on". A guard that fails open and quiet is worse than no guard.
- **resolveModel returned any | string.** The success type was any and the failure channel was a human-readable message, so all seven call sites discriminated with typeof x === "string" — the model-versus-error distinction existed only as prose.
- **A failure and a legitimate zero were the same value.** getSessionTokens returned 0 both for "nothing spent yet" and for "reading the session's stats threw".

## What changed

neverthrow is now the internal currency for the model path, the JSON-Schema compiler and the usage readers. The one idea: **Result is internal, and every pi-facing callback converts once at its own edge.** That is not a style preference — pi's tool execute has no structured error channel (it marks a call failed only when execute throws; a returned isError flag is ignored), pi.on handlers swallow throws into a toast and discard return values, and a throwing pi.events handler never replies at all. So each conversion is explicit and single-point: the RPC spawner throws the message into its envelope, the Agent tool and the nested tools render it as tool text, the workflow host returns its own ok/error shape, and the schedule path keeps falling back silently.

Consequences of that one idea:

- src/model/errors.ts is the resolution failure domain — one NOT_FOUND variant built by autoTag (src/lib/result.ts), so the object key *is* the runtime code and an entry cannot drift from it. Its message is quoted verbatim in docs/rpc.md and is unchanged.
- resolveModel is an .orElse chain (exact then fuzzy then provider fallback), not a safeTry generator: each attempt recovers from the previous one's miss, and safeTry short-circuits on the first Err and would never reach the second attempt. compileJsonSchema is the opposite case — five sequential guards where short-circuiting is exactly right — and is one safeTry pipeline.
- The scopeModels policy becomes a ModelScope instance built in index.ts and threaded through the existing deps slices, replacing a module-level boolean with two writers (the settings applier and the /agents Settings toggle). Leaves receive it through their deps slice, never by importing a singleton.
- readEnabledModels returns a Result: Ok(undefined) is "no allowlist" (absent is not an error), while Err names the file that exists and cannot be parsed. The scope check turns that into the warn channel the policy already had.
- CompiledSchema.check returns Result of true or string — the compiler already had an ok-union, so this is a swap rather than an addition.

## Related work

| Number | Title | State | Relation |
|---|---|---|---|
| #5, #6, #7 | the src/ domain-layout reshape | merged | parent work; this continues the handover it left |
| #8 | rpc citation re-verification | merged | its citations are corrected here for lines this PR moved |

## Behavior and compatibility

No user-facing change: tool names, settings keys, lifecycle events, RPC channels, error text and on-disk paths are all untouched. Two docs/rpc.md error-table citations are updated because the lines moved — the model-not-found text is now in src/model/errors.ts:24, the scope message in src/model/model-scope.ts:124.

One deliberate behavior change: an unparseable settings file now warns on the next spawn, naming the file. Previously that condition silently passed every model. Enforcement is still fail-open — refusing every spawn would turn a typo in one JSON file into a dead extension.

Removed: getSessionTokens, whose only consumer in the repo was its own test.

neverthrow ^8 is a new runtime dependency. pi loads this extension as TypeScript source (pi.extensions: ./src/index.ts) with no bundler, so a runtime dependency resolves exactly like the existing croner / nanoid / typebox ones.

## Performance

Not measured — no benchmark was run for this PR. From the diff, the only hot-path note is that resolveModel's success path is unchanged in shape (one Set build plus the scoring loop; the failure list stays lazy, as before) and a Result adds one small object on failure only.

## Testing

    npm run lint       # biome: 0 errors (5 pre-existing warnings under test/perf/)
    npm run typecheck  # tsc --noEmit: 0 errors
    npm run test       # 2185 passed | 24 failed | 7 skipped (2216) across 117 files

The 24 failures are pre-existing machine state, not this PR: the same 7 files with the same per-file counts on pristine master. The machine's ~/.pi/agent/agents/ shadows the repo's built-in agent types, so tests that spawn general-purpose get "Unknown or disabled agent type". Verified by running those files in a throwaway git worktree of 6b98c1b — 24 failed / 189 passed — and diffing the failure sets, which match exactly.

New coverage:

- test/neverthrow-fence.test.ts locks the architectural invariant: a catch is allowed only in the enumerated adapters of the pilot modules, named by the enclosing declaration. A new try/catch in those files fails a test rather than passing review.
- test/model-scope.test.ts gains the corrupt-settings case, the no-available-models no-op, duplicate-pattern listing, and per-test instances instead of a saved-and-restored module global.
- test/enabled-models.test.ts is rewritten for the Result API, asserting the channel before the payload and pinning which file a corrupt read reports.

Not covered: no benchmark (above); no rung-3 TUI drive, since no render path changed — the one new user-visible string (the corrupt-file warning) is asserted through the verdict rather than a rendered frame.


gjuoun:
Opened against the wrong base by mistake — this belongs on the fork (gjuoun/pi-subagents). Closing; sorry for the noise.

PR #335, opened 2026-09-19
fix: survive session replacement when a flag-launched workflow finishes

ajaytravel:
A `--subagents-workflow-file` run is detached in TUI/RPC, so its completion fires whenever it fires — possibly after `/new`, `/fork`, `/resume` or `/reload` replaced the session. The captured `pi` handle is stale by then and `appendEntry` throws. Nothing caught it, so pi exited 1, losing the run's result and killing any child still in flight.

**Repro** (pi 0.85.1, master `e955e29`): `pi --mode rpc --subagents-workflow-file=slow.js` where the script's child sleeps 25s; send `{"type":"new_session"}` while it runs; wait.

```
error: This extension ctx is stale after session replacement or reload...
    at appendEntry
    at <anonymous> (src/index.ts:2719:21)
```

**Fix:** delivery is best-effort — try/catch around `appendEntry`/`sendMessage` and the widget/fleet repaint, with a `console.warn` on the loss. Not `report()`, since `ctx.ui` is stale for the same reason. `withSession` isn't available here: it belongs to whoever called `newSession`, which isn't this extension. The task stays in `workflowTasks` either way.

Test added in `workflow-tool.test.ts`; it fails on master with the unhandled rejection vitest reports. Lint/typecheck/test/build all pass.


ajaytravel:

PR #336, opened 2026-09-19
fix: await a flag-launched workflow in one-shot headless modes

ajaytravel:
Stacked on #335 (same block of `runWorkflowFlag`); the second commit is the change here.

Detaching the `--subagents-workflow-file` run is right for TUI and RPC, where `session_start` must not block startup. Under `pi -p` (mode `print`, or `json` with `--mode json`) it's wrong: the process runs one turn and exits, so the detached run races teardown and usually loses — dying part-finished and aborting its in-flight children. There's also no later turn for its `deliverAs: "nextTurn"` message to land on.

**Repro** (pi 0.85.1): `pi --mode json --subagents-workflow-file=nest2.js -p 'Reply OK.'` with a two-level workflow. The main turn answered in ~10s and pi exited while the grandchild was still generating; the child session recorded `stopReason: "aborted"`. Same script via RPC completed.

**Fix:** `runWorkflowFlag` becomes async and is awaited from `session_start` only when `ctx.mode` is `print` or `json` — in those modes the workflow *is* the invocation. TUI and RPC keep the detached behaviour. Verified live: `pi -p --subagents-workflow-file=tiny.js "reply with the secret field"` now returns the workflow's value, so the `nextTurn` message lands on the one turn.

Tests added for both modes (no `waitFor` — the run must have settled when `session_start` resolves); they fail on master. `docs/workflows.md` updated to say which modes detach and which await. Lint/typecheck/test/build all pass.


ajaytravel:

PR #337, opened 2026-09-19
fix: honour an agent's `isolated` in SubagentWorkflow spawns

ajaytravel:
The `Agent` tool reads `isolated` off the agent definition and passes it to the manager as a spawn option. The workflow host resolves the same config (for model precedence) but never forwarded the flag, so an agent file's `isolated: true` was silently dropped for every child a script spawned.

That inverts the flag's purpose. `isolated` draws a hermetic boundary — no extensions, no skills, no `ext:` selectors — and the agents given it are usually the read-only ones. Dropped, they loaded the full host extension set.

**Observed** (pi 0.85.1, master `e955e29`): asked the same read-only `scout` definition to enumerate its tools down both paths. Via `SubagentWorkflow` it listed `web_search`, `mcpScript` and six `mcp__*` tools; via `Agent` it correctly had none. The builtin allowlist applied in both cases, so only the extension half leaked — which is what made it quiet.

**Fix:** one line in `workflow/host.ts` forwarding `config?.isolated` when set (undefined stays undefined, so the runner's default still decides).

New `test/workflow-isolated.test.ts` covers `true`, explicit `false`, and unset; the first two fail on master. Lint/typecheck/test/build all pass.


ajaytravel:

